PR #115, opened 2015-06-01
tomcat 8 prep

danfruehauf:
Still leaves us at tomcat7 and openjdk7 - however this is some preliminary work before we can embed the following in edge and systest, then in all:
```
    "tomcat": {
        "fine_version": "8.0.32",
        "pkg_url": "https://archive.apache.org/dist/tomcat/tomcat-8/v8.0.32/bin/apache-tomcat-8.0.32.tar.gz",
        "pkg_checksum": "7e23260f2481aca88f89838e91cb9ff00548a28ba5a19a88ff99388c7ee9a9b8"
    },
    "java": {
        "jdk_version": 8
    },
```

The annoying part will come when we need to rename the `tomcat7` user to `tomcat` and `/mnt/ebs/tomcat7` to `/mnt/ebs/tomcat`. I think that doing that manually can be faster - or not manually, with a script - just not with chef. Renaming the user will require killing all the tomcat containers because otherwise Linux will prohibit us from doing that.

Something along the lines of this will probably solve the problem:
```
#!/bin/bash

# stop all tomcats
for tomcat in /etc/init.d/tomcat_*; do $tomcat stop; done

# move homedir (doesn't exist anyway :)
usermod -d /home/tomcat tomcat7

# rename user
usermod -l tomcat tomcat7

# rename group
groupmod -n tomcat tomcat7

# move tomcat home
mv /mnt/ebs/tomcat7 /mnt/ebs/tomcat

# fix init scripts - chef will do it anyway, but try to minimize downtime
for tomcat in /etc/init.d/tomcat_*; do sed -i -e 's/tomcat7/tomcat/g' $tomcat; done

# start tomcats again
for tomcat in /etc/init.d/tomcat_*; do $tomcat start; done
```

### DIFF
--- a/cookbooks/imos_webapps/recipes/generic_webapp.rb
+++ b/cookbooks/imos_webapps/recipes/generic_webapp.rb
@@ -38,7 +38,7 @@ if node['webapps'] && node['webapps']['instances']
     instance_aliases           = instance['aliases']
     instance_tomcat_port       = instance['port']
     instance_base_directory    = "#{node['tomcat']['base']}/#{instance_name}"
-    instance_service_name      = "tomcat#{node["tomcat"]["version"]}_#{instance_name}"
+    instance_service_name      = "tomcat_#{instance_name}"
     instance_data_directory    = instance_parameters['data_dir'] || "#{instance_base_directory}/data/#{instance_name}"
     instance_apps              = instance['apps']
     instance_httpd_rules       = instance['httpd_rules']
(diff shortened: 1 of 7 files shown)


danfruehauf:
I'm deploying that tonight. It's been a while and it is OK to be deployed as it doesn't affect anything (tested in vagrant).

danfruehauf:
No explosions.

Issue #241, opened 2016-01-22
imos_po: Double handling of pipeline files


danfruehauf:
## How to reproduce:
 * Build an incoming handler that takes a long time to process (insert a `sleep 60` for instance)
 * Drop a file in the incoming directory
 * Restart all celery processes:
```
$ sudo supervisorctl restart all
```
 * Alternatively, just restart the handler you are using and `inotify_po`, like:
```
$ sudo supervisorctl restart inotify_po
$ sudo supervisorctl restart celery_po_ACORN
```

## What happens
 * File is processed twice
   * File already started processing
   * Triggering a restart for `inotify_po` pushes the file again to the queue
   * Restarting `celery_po_ACORN` detaches from the currently running process and consumes the file again from the queue

## What should happen
 * File is processed only once

## Why is that a problem?
The problem is that provisioning a machine will cause a restart of all celery processes, which can result in this awkward situation. The situation is especially problematic for files that take a long time to process - ARGO for instance

While there are a few ways of approaching this bug, I thought it's best to first communicate it and then propose solutions.

danfruehauf:
This is a problem also for ANMN_NRS, which can receive large zip files that can take a while to process.

Probably need to look at solving this sooner than later.

mhidas:
@danfruehauf Not just ANMN_NRS, but most of the other ANMN pipelines and possibly also ABOS.

danfruehauf:
I'm going to thoroughly test it too now.

danfruehauf:
Closed by #255.

PR #255, opened 2016-02-09
imos_po: Watch exec wrapper sandboxing

danfruehauf:
Move file to temporary directory before processing it. This will
prevent double processing of files like described in #241.

Also with:
https://github.com/aodn/chef-private/pull/1776


danfruehauf:
I gave it the testing I want.

@mhidas Did you get to test it?

@lbesnard FYI

mhidas:
@danfruehauf Yep, didn't cause any problems.

danfruehauf:
Feel free to hit the merge then, and also on aodn/chef-private#1776

PR #256, opened 2016-02-26
ADD databags for AATAMS SATTAG NRT

lbesnard:


private-sample/data_bags/talend/aatams_sattag_nrt.json
```diff
@@ -0,0 +1,15 @@
+{
+    "id": "aatams_sattag_nrt",
+    "artifact_filename": "aatams_sattag_nrt_harvester_Latest.zip",
+    "params": {
+        "Metadata_UUID": "b4e742f9-58e5-46b8-b365-15430ea45599",
+        "Metadata_SpatialTable": "aatams_sattag_nrt_profile_map",
+        "Metadata_SpatialResolution": "2",
+        "metadata_file": "['imos_po']['data_services']['dir']/AATAMS/aatams_sattag_metadata.csv"
```
lbesnard: @danfruehauf , xav uses this file in his harvester. Could not test this on PO. Only did testing with full path to file

danfruehauf: That should be alright. Though, why not use `Metadata_File` - a bit cleaner.

lbesnard: I asked xav to harvest the information from the netcdf file now. So he'll have to modify the harvester, and I'll get rid of this line. But for now it will stay as is as he's going tomorrow I think

PR #257, opened 2016-02-28
README on how to add swap space

danfruehauf:


doc/README.swap.md
```diff
@@ -0,0 +1,15 @@
+Swap Space
+==========
+
+Using swap is not recommended, however the nature of java apps is to consume
+a lot of memory even when doing nothing. To create swap space on EBS, use the
```
jonescc: This seems a little unfair - last comments on that issue point to ncks.  Its not the nature of java apps to use lots of memory - just badly written ones.  There's normally an underlying cause which should be looked at not worked around.

danfruehauf: True. Although ncks runs "on demand", whereas java is hogging memory constantly.

I still think that's applicable to the nature of our webapps, regardless of ncks.

danfruehauf: If this is controversial, I can just remove that comment. However I felt like there should be an explanation as to "why we need swap", and that is because "java is using a lot of memory even when doing nothing".

jonescc: I don't think we should be rubbishing java - its not a java issue - there's are issues with memory management in some of the applications we are using.

danfruehauf:
@jonescc Java close to your heart. Maybe I should have said *tomcat*, more accurate. Either way, the comment about that is removed. What I just wanted to do is document how to add swap space.

PR #265, opened 2016-03-04
imos_po: Bulk async upload option

danfruehauf:
Upload many files asynchronously by piping them to
`async-upload.py`.

Example usage for a single file:
```
$ echo "SRC DST" | async-upload.py
$ echo "/tmp/IMOS_ACORN_V_20100716T060000Z_BONC_FV00_sea-state.n IMOS/ACORN/vector/BONC/2010/07/16/IMOS_ACORN_V_20100716T060000Z_BONC_FV00_sea-state.nc" | async-upload.py
```

Uploading multiple files:
```
$ find . -type f | cut -c3- | awk -v pwd=$PWD '{print pwd"/"$1 " " "IMOS/"$1}'
/tmp/ABOS_SOFS/ff/d/e/file2 IMOS/d/e/file2
/tmp/ABOS_SOFS/ff/file3 IMOS/file3
/tmp/ABOS_SOFS/ff/a/b/c/file1 IMOS/a/b/c/file1

$ find . -type f | cut -c3- | awk -v pwd=$PWD '{print pwd"/"$1 " " "IMOS/"$1}' | async-upload.py
```

### DIFF
--- a/cookbooks/imos_po/attributes/data_services.rb
+++ b/cookbooks/imos_po/attributes/data_services.rb
@@ -90,3 +90,5 @@ default['imos_po']['data_services']['create_watched_directories'] = false
 
 default['imos_po']['data_services']['credentials_pattern'] = "*"
 default['imos_po']['data_services']['credentials_prefix'] = "IMOS_PO_CREDS"
+
+default['imos_po']['data_services']['async_upload']['max_tasks'] = 4
--- a/cookbooks/imos_po/libraries/watch_jobs.rb
+++ b/cookbooks/imos_po/libraries/watch_jobs.rb
@@ -26,4 +26,13 @@ class Chef::Recipe::WatchJobs
     return watchlists
   end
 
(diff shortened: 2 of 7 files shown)


danfruehauf:
@lbesnard Would be very useful for uploading whole AUV campaigns.

@mhidas Would be useful for uploading large amounts of data such as passive acoustics.

This will work on the PO box as expected - moving files to `/s3/imos-data`

danfruehauf:
I forgot to mention that this is the technique I used to upload the SRS data - and that was quick. By default, it'll use up to 4 concurrent threads to upload the data and the queuing of the files should be super quick and bound only by the time it takes `find` to iterate on the filesystem.

mhidas:
Obviously this is upload without indexing, right?

danfruehauf:
> Obviously this is upload without indexing, right?

Yes. It is supposed to assist with bulk uploading in the future. It's a "power user" feature.

PR #266, opened 2016-03-06
imos_po: Create watches on first provision

### LINKED ISSUE #240: New watches not deployed on first chef provision
How to reproduce:
 * `vagrant up 10-aws-syd`
 * Add a new watch in `data-services/watch.d`
 * `vagrant provision 10-aws-syd`

What happens:
 * Machine provisions, but the new watch is not defined in `/mnt/ebs/data-services/celery/inotify-config.py`

What should happen:
 * Watch is defined in `/mnt/ebs/data-services/celery/inotify-config.py`

This is due to the watches being generated before the git repository is being updated/pulled. It does not happen on the po box. On the po box the `data-services` git repository is not "managed" and does not get updated/pulled.

danfruehauf:
The problem this PR solves is described in #240.

The fix is rather awkward, but works properly. Any related
template would subscribe to be notified after git[data_services]
is updated, however it would evaluate the list of watches before
the git repository was updated as a template variables.
Eventually when the template is evaluated, it would use a stale
watch list which consists of watches *before* git[data_services]
was updated. The template however would be evaluated after
git[data_services] was updated.

The fix moved the evaluation of the watch list to the template
itself. This causes the list of watches to be evaluated *after*
git[data_services] was updated, thus resulting in the right list
of watches used by the template.

Fixes #240

### DIFF
--- a/cookbooks/imos_po/recipes/watches.rb
+++ b/cookbooks/imos_po/recipes/watches.rb
@@ -94,7 +94,7 @@ if node['imos_po']['data_services']['watches']
     source     "tasks.py.erb"
     variables  ({
       :celery_config     => ::File.basename(celery_config),
-      :watchlists        => Chef::Recipe::WatchJobs.get_watches(data_services_watch_dir),
+      :watch_dir         => data_services_watch_dir,
       :data_services_dir => data_services_dir
     })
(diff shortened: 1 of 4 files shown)


danfruehauf:
Don't want this to rot.

PR #269, opened 2016-03-08
imos_po: s3lsv - List versions of file on S3

danfruehauf:
Explore previous versions of objects using `s3lsv`.

Basic usage:
```
$ s3lsv
usage: s3-lsv.py [-h] [-a] -b BUCKET -k KEY [-o OUTPUT] [-v VERSION]
s3-lsv.py: error: argument -b/--bucket is required

$ s3lsv -a -b imos-data -k IMOS/
IMOS/AATAMS/
IMOS/ABOS/
...
IMOS/SRS/

$ s3lsv -a -b imos-data -k IMOS/ACORN/radial/TAN/2016/01/01/IMOS_ACORN_RV_20160101T235000Z_TAN_FV00_radial.nc
na5gI3LrPUyyfZ3JsHDFDV3WTAmFt2n5 2016-01-01T23:55:13.000Z IMOS/ACORN/radial/TAN/2016/01/01/IMOS_ACORN_RV_20160101T235000Z_TAN_FV00_radial.nc

$ s3lsv -a -b imos-data -k IMOS/ACORN/radial/TAN/2016/01/01/IMOS_ACORN_RV_20160101T235000Z_TAN_FV00_radial.nc -o /tmp/file.nc -v na5gI3LrPUyyfZ3JsHDFDV3WTAmFt2n5
$ ls -l /tmp/file.nc
-rw-rw-r-- 1 vagrant vagrant 263728 Jan  2 10:55 /tmp/file.nc
```

### DIFF
--- a/cookbooks/imos_po/attributes/data_services.rb
+++ b/cookbooks/imos_po/attributes/data_services.rb
@@ -59,6 +59,7 @@ default['imos_po']['data_services']['packages'] = [
   'libnetcdf-dev',
   'mdbtools',
   'python-beautifulsoup',
+  'python-boto',
   'python-dev',
   'python-pip',
   'python-psycopg2',
--- a/cookbooks/imos_po/files/default/s3lsv
+++ b/cookbooks/imos_po/files/default/s3lsv
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+
+import argparse
+import boto
+from boto.s3.prefix import Prefix
+
+def get_object_with_version(bucket, key, version, f):
+    k = bucket.get_key(key, version_id=version)
+    k.get_contents_to_filename(f)
+
+def list_object_versions(bucket, key):
+    for k in bucket.list_versions(key, delimiter='/'):
+        if type(k) is Prefix:
+            print k.name
+        else:
+            print "%s %s %s" % (k.version_id, k.last_modified, k.name)
+
+if __name__=='__main__':
+    parser = argparse.ArgumentParser()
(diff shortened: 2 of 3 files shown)


danfruehauf:
@mhidas @lbesnard @ggalibert @pblain @bpasquer @xhoenner FYI

If anyone can come up with a better name than `s3lsv` - please suggest it. I chose s3lsv because it's like s3ls, but with version support.

Perhaps it is something I should try and contribute upstream to `s3cmd`.

mhidas:
@danfruehauf Will the output indicate which is the *current* version?

danfruehauf:
@mhidas That should give it to you:
```
$ s3lsv -a -b BUCKET -k KEY | sort -k2 | tail -1
```

That'll sort it by date.

mhidas:
Could we make it so the output is always sorted by date? Or rather by key, then date, if `KEY` can match multiple objects. I think that makes sense.

danfruehauf:
> Could we make it so the output is always sorted by date? Or rather by key, then date, if KEY can match multiple objects. I think that makes sense.

Could do. Latest is last or first?

danfruehauf:
> Could do. Latest is last or first?

@mhidas Looks like it'll sort it by default. Latest is first.

mhidas:
:+1:

danfruehauf:
Feel free to merge it then. It's a utility that will be installed on 10-aws-syd. It will not really be usable on the po box.

PR #271, opened 2016-03-10
ANFOG trigger

danfruehauf:


danfruehauf:
Added another commit here. This removes the declarations of the OPENDAP and PUBLIC dir. After ANFOG is merged, nothing else should be using those...

danfruehauf:
@mhidas @lbesnard @bpasquer FYI no more `$OPENDAP_DIR` and `$PUBLIC_DIR` environment variables - for obvious reasons.

PR #277, opened 2016-03-21
add gdal python support

lbesnard:


danfruehauf:
@lbesnard I need to revert it because it is conflicting with postgis 2.1 and destroying a few essential things.

Please plan with devs on how to get it in.

PR #283, opened 2016-04-11
Updated keyserver url

bhasin85:
Please test it in your environment before the merge.

### DIFF
--- a/cookbooks/imos_postgresql/recipes/official_postgresql.rb
+++ b/cookbooks/imos_postgresql/recipes/official_postgresql.rb
@@ -22,7 +22,7 @@ apt_repository 'official_postgres' do
   uri 'http://apt.postgresql.org/pub/repos/apt/'
   distribution "#{node['lsb']['codename']}-pgdg"
   components ['main']
-  keyserver 'keyserver.ubuntu.com'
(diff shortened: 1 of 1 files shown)


cookbooks/imos_postgresql/recipes/official_postgresql.rb
```diff
@@ -22,7 +22,7 @@
   uri 'http://apt.postgresql.org/pub/repos/apt/'
   distribution "#{node['lsb']['codename']}-pgdg"
   components ['main']
-  keyserver 'keyserver.ubuntu.com'
+  keyserver 'hkp://keyserver.ubuntu.com:80'
```
pmbohm: If this is causing problems, would it also be causing the same problem in `..../chef/cookbooks/imos_java/recipes/default.rb`
?

lwgordonimos:
Can this now be deleted? I'm not aware of any issues currently, and there was a change to update an expired key which may have been the cause of this original issue, https://github.com/aodn/chef/pull/315

lwgordonimos:
@bhasin85 ?

bhasin85:
Look like it got fixed with the other pull request you mentioned. I will close this one now.

PR #291, opened 2016-04-26
Logos for Geonetwork using HTTPS

pmbohm:


pmbohm:
Fixes an issue where the whole page is not reported by the browser as being secure

PR #302, opened 2016-05-19
Chef changes for running Portal in root context

lwgordonimos:
* replace several instances of "Hash.first" with named indices
* add "default_app" attribute to Tomcat instance config item, and use this to inject an "is_default_app" flag into each app config item to allow it to set it's configuration accordingly. This enforces that only one app per instance may be assigned as default
* For default app deployment, deploy to ROOT Tomcat context (i.e. deploy as ROOT.war)


anguss00:
👍

PR #306, opened 2016-05-20
Add tablespace creation definition, and add to Postgres recipe

lwgordonimos:
Reads "tablespaces" list from data bag, creates them if they don't exist. This allows the postgresql.conf file to use one or more of the defined tablespaces in the "temp_tablespaces" configuration item.

See private node configuration file for node configuration to enable this. Note that this method makes the same assumption as the main Postgres data directory, which is that the underlying volume *already exists* and the filesystem has been created (i.e with mkfs) and is available at the device node named in the 'mounts' array. This does means that it will not work in one step on a "from scratch" install outside of vagrant, however I believe this is the status quo anyway.


julian1:
Nice!

PR #314, opened 2016-06-06
Add GDAL w/Python recipe to imos_po cookbook

lwgordonimos:


danfruehauf:
G'day. I think that before opting to the `./configure && make && make install && ldconfig` option, have you looked at building a package and installing it with `dpkg`? Jenkins can build it, then the chef recipe can just download & install it. `./configure && make && make install` is probably the dirtiest solution to install something on *nix.

lwgordonimos:
Hey. Yes, a DEB was my first intention. This is proof of concept while confirming a few things with POs (in a non-vagrant environment) and not intended for production "as is". Also mostly a practice recipe for me ;)

I will say that for libraries which may conflict with system packages (e..g libgdal) that you can easily get in a DPKG dependency hell trying to conflict/supersede the core packages, so a self-contained build off in /usr/local is not always a dirty option per se, as long as it doesn't mix with system packages (e.g. in /usr/lib). 

Agree regarding precompilation before deployment however, not a fan of compiling anything on prod servers (or even having compilers on prod systems to be honest).

lwgordonimos:
Closed because this actually shouldn't have been a pull request for testing

lbesnard:
having a PR helps you to create a discussion at least

lwgordonimos:
Ok we'll keep discussion here for this one with the condition that this code doesn't get merged ;)

lbesnard:
thks. Also, not a bad thing to add a short 'not to merge' type of txt at the top just in case. So you're sure there is no confusion with other people

lwgordonimos:
Refer https://github.com/aodn/chef/pull/319 for Jenkins/DEB based solution now that it has been tested on 14

lbesnard:
@lwgordonimos 
I have a similar issue on aws10 as we first had on nsp14.  ```Unable to load PROJ.4 library (libproj.so)```

lwgordonimos:
This should be resolved. The 'libproj.so' symlink for some reason resides in the -dev package rather than the main package.

PR #318, opened 2016-06-09
Installing IMOS Compliance Checker Plugin

bhasin85:
**For review purposes only! @anguss00 

Do not merge.**


cookbooks/imos_po/attributes/netcdf_checker.rb
```diff
@@ -1,4 +1,10 @@
...
+default['imos_po']['netcdf_checker']['dir'] = "/var/lib/netcdf-checker"
+default['imos_po']['netcdf_checker']['repo'] = "https://github.com/ioos/compliance-checker.git"
+default['imos_po']['netcdf_checker']['branch'] = "master"
 default['imos_po']['netcdf_checker']['executable'] = "/usr/local/bin/netcdf-checker"
+
+# IMOS Compliance Checker Plugin
+default['imos_po']['cc_plugin_imos']['dir'] = "/var/lib/imos_checker"
+default['imos_po']['cc_plugin_imos']['path'] = "/var/lib/imos_checker/lib/cc_plugin_imos"
+default['imos_po']['cc_plugin_imos']['repo'] = "git@github.com:aodn/data-services.git"
+default['imos_po']['cc_plugin_imos']['branch'] = "imos_checker_plugin"
```
lbesnard: by the look of it, the data services repo seems to be checked out out using a specific ```imos_checker_plugin``` branch. Why this ? Why not doing a symbolic link between the imos checker file stored in $DATA_SERVICES_DIR directly into the IOOS checker path, especially since the data service is already cloned in nodes where the imos plugin will be ? Sorry if I misunderstood

bhasin85: @lbesnard  You are right that the data service is already cloned and re-cloning it does not make much sense. I created this PR just to get some feedback from @anguss00 but forgot to mention it in the description. Sorry for the confusion.

danfruehauf: @lbesnard has a point. His solution is:
 * 4 less attributes (without destroying the indent and the newline in end of file)
 * 10 less lines of code
 * Significantly less ugly and less clutter to cleanup in the future
 * Just as flexible

PR #341, opened 2016-07-06
Deploy compliance_checker python eggs built using jenkins

anguss00:
IMOS plugin installed using command due to bug in chef 12.04
@lwgordonimos


lwgordonimos:
@anguss00 updated

PR #343, opened 2016-07-11
Update PO box compliance checker version

anguss00:
@mhidas @ggalibert @lbesnard this updates the version on a fresh po box, you potentially need to uninstall the old plugin and compliance checker if updating an existing box.

### DIFF
--- a/cookbooks/imos_po/attributes/netcdf_checker.rb
+++ b/cookbooks/imos_po/attributes/netcdf_checker.rb
@@ -1,6 +1,4 @@
 default['imos_po']['netcdf_checker']['dir']           = "/usr/local/bin"
-default['imos_po']['netcdf_checker']['repo']          = "https://github.com/ioos/compliance-checker.git"
-default['imos_po']['netcdf_checker']['branch']        = "master"
 default['imos_po']['netcdf_checker']['executable']    = "/usr/local/bin/netcdf-checker"
 default['imos_po']['netcdf_checker']['version'] = '2.2.0'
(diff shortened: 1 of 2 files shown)


lwgordonimos:
@anguss00 Actually this whole node can be removed, since it's prebuilt in the EGG, and nothing uses that git info any more:

https://github.com/aodn/chef/commit/7736a03db19196f0b94a1c8a246cf99c3b631768

anguss00:
Only thing is, the node allows the base checker branch to be set. The default set in attributes is IOOS main, while we have the aodn fork overwritten. I agree though, it's ripe for just configuring on the attribute

lwgordonimos:
See https://github.com/aodn/data-services/pull/523#issuecomment-231636662

cookbooks/imos_po/attributes/netcdf_checker.rb
```diff
@@ -1,5 +1,4 @@
 default['imos_po']['netcdf_checker']['dir']           = "/usr/local/bin"
-default['imos_po']['netcdf_checker']['repo']          = "https://github.com/ioos/compliance-checker.git"
 default['imos_po']['netcdf_checker']['branch']        = "master"
```
lwgordonimos: This ['repo'] line can also be removed as it's now unused in the client recipe. The ['dir'] is only used by the manual symlink, but I don't think it's worth doing any work to remove that given the future plans for this.

anguss00: Ah yes, should definitely remove the "branch" line. Will amend

anguss00:
Doneskis @lwgordonimos

PR #346, opened 2016-07-14
ANMN metadata harvester to index ABOS files too

mhidas:


mhidas:
This can be merged now.

PR #367, opened 2016-07-26
anmn_acoustics talend job trigger

mhidas:


mhidas:
@ggalibert Would you mind merging this too, just to make the PO Box consistent with what is already in production (see https://github.com/aodn/chef-private/issues/2008)

mhidas:
@ggalibert Thanks!

Issue #366, opened 2016-07-25
async-upload doesn't work on 10-aws-syd


mhidas:
The `asynd-upload.py` script allows us to bulk upload a bunch of files directly to S3 without indexing (see #265 ). It used to work fine, and still does on 14-nsp-mel (uploading to the *imos-test-data* bucket), but I tried to use it on 10-aws-syd last week and nothing has been uploaded. No error messages are printed. Maybe these uploads are being added to a queue, but they are not being processed.

There's also no easy way to check what this command is doing (https://github.com/aodn/internal-discussions/issues/260)

lwgordonimos:
@mhidas Looking now...

mhidas:
@pblain

mhidas:
Thanks @lwgordonimos!
I think the conclusion was that my uploads were failing because the projectofficer user (or is it root?) had no read access to the files I was trying to upload.
In any case, having an easily accessible error log at /mnt/ebs/log/data-services/async_upload.log will be very handy in the future.

Will that log get rotated daily? Or deleted?

lwgordonimos:
@mhidas The logging is still the default values, which looks to clean them up whenever the supervisor daemon is restarted.

mhidas:
How often does that happen?

lwgordonimos:
For example, the current logs go back three weeks, but we should probably be more explicit to have a predictable retention period at least for this log, since it's the only record that this command has been run (other than the existence or absence of the files in S3)!

mhidas:
Yes, let's do that!

lwgordonimos:
@mhidas FYI, there are two PRs which will prevent the logs being clobbered on every service restart, and moved them to the data services logging area as well.

Accordingly, I've reduced the retention a bit on the 'normal' automatic queues (as opposed to the interactive async_upload) to prevent too much storage wastage, since the 'nocleanup' option is global to supervisord and not able to be set per-queue.

https://github.com/aodn/chef/pull/377
https://github.com/aodn/chef-private/pull/2011

mhidas:
Cool, thanks!

PR #369, opened 2016-07-26
Add check to async_upload to avoid silent s3cmd failure

lwgordonimos:


lwgordonimos:
Fixes https://github.com/aodn/chef/issues/366

The s3cmd has *very* inadequate handling of errors, and so returns 0 when at least 2 (very likely more) scenarios occur:
1) the celery user doesn't have read access to the file
...
1000000) the file doesn't exist
etc.

Ideally this will be uploaded using native Python AWS (i.e. boto), but in the mean time this adds a simple test to cause the task to fail and at least log the fact that it has failed.

e.g.

```bash
14-nsp-mel:/var/log/supervisor# s3cmd --config=/mnt/ebs/data-services/s3cfg sync --no-preserve /root/NOFILEHERE s3://imos-test-data/NOFILEHERE
14-nsp-mel:/var/log/supervisor# echo $?
0
```

julian1:
Is there a python S3cmd equivalent?

Would it be better to replace s3cmd wholesale?

julian1:
Also, I believe that [aws-cli](https://aws.amazon.com/cli/) the official cli client that will have AWS support. While s3cmd is a third party app. It would be interesting to know if the error-handling is better with the official client.

lwgordonimos:
@julian1 I briefly referenced boto which is the Python AWS library, which would be preferable to shelling out to anything since it can be done in native Python.

We should definitely change it to either of the more robust tools (cli for bash or boto lib for Python) but at a cursory glance a heck of a lot of things are built around s3cmd specifically so would need substantial testing before migrating, hence this very conservative check in the interim. 

We would need to decide, as a rule, that we consider boto as "safe" to handle file uploads. I suspect it will be fine, since it is the AWS endorsed library (https://aws.amazon.com/sdk-for-python/) but didn't want to make such a change here without it being well tested first and also done with the wider scope in mind.

cookbooks/imos_po/templates/default/tasks.py.erb
```diff
@@ -40,15 +42,17 @@ s3cmd_quoted = s3cmd_parts.map{ |s| "\"#{s}\""}.join(', ')
...
 def async_upload(src_path, dst_path):
+    if not os.access(src_path, os.R_OK):
+        raise Exception("ERROR uploading '{0}'. The file must exist and user '{1}' must have read access to it.".format(src_path, getpass.getuser()))
     command = [
         <%= s3cmd_quoted %>,
         "sync",
         "--no-preserve",
         src_path,
-        path.join("<%= node['imos_po']['s3']['bucket'] %>", dst_path)
+        os.path.join("<%= node['imos_po']['s3']['bucket'] %>", dst_path)
```
julian1: Can't believe we templatized the python to achieve the necessary level of expressibility. And hanging a dependency like our entire chef-framework on this in order to generate runnable code. 

Should be re-factored and made config/data driven.

PR #371, opened 2016-07-27
Remove use of unused node attributes

jonescc:
Could not find any references to these node attributes elsewhere - we are currently using config on tomcat applications in node files.

Ensure no one else has to spend 1/2 hour working this out.


jonescc:
Some history - we used to have a cookbook for each web application and we used to use attributes such as those being referenced here to configure them.   We don't do that anymore.

PR #372, opened 2016-07-27
Aatams, fix image and static download links

julian1:


cookbooks/imos_webapps/templates/default/aatams/aatams.groovy.erb
```diff
@@ -20,7 +20,7 @@ grails {
     }
 }
 
-grails.serverURL  = "<%= @custom_parameters['https'] ? 'https' : 'http' %>://<%= @custom_parameters['vhost'] %>/<%= @custom_parameters['name'] %>"
```
lwgordonimos: The one comment I have is that the grails.serverURL should also cater to the scenario where it is *not* the default app, which is a flag derived from the node file.

See the change to grails.serverURL in _cookbooks/imos_webapps/templates/default/portal/Portal.groovy.erb_ in this commit, which makes the context optional rather than unconditionally removing it:
https://github.com/aodn/chef/commit/01a946fefd29cbd6bfd9fc0304a196bba590b5e3#diff-a383edddf09534ed3231a493834c4217

julian1: I'll fix

julian1:
updated

PR #373, opened 2016-07-27
Move location of async_upload log to be accessible to POs

lwgordonimos:
Since this queue is as the result of an interactive command executed by POs, it makes sense that this log is available to them to verify the success/failure of their file uploads. 

Due to a supervisord bug, when using the default "AUTO" location for log files, these are created with root-only access, so explicitly defining the path is required for permissions reasons. In addition, it makes sense that the log is in the data services log dir, as it is more intuitive for POs to find it there with the other logs than in /var/log/supervisor.


lwgordonimos:
@mhidas This in association with https://github.com/aodn/chef/pull/369 should fully resolve https://github.com/aodn/chef/issues/366

anguss00:
Just checking out in vagrant, then will merginate

Issue #370, opened 2016-07-26
Local artifact deployment now requires md5 file


jonescc:
I make use of local artifact deployment extensively for testing modified code in vagrant.

e.g.

                "apps": [
                    {
                        "name":               "geoserver",
                        "artifact":           "/mnt/source-code/geoserver-build/src/main/target/geoserver-1.0.0-imos.war",
                        "custom_def":         "geoserver",
                        "data_bag":           "geoserver_123",
                        "jndis":              [ "harvest_read_123_1_nec_hob" ],

This recently stopped working for me - seems an md5 file is required which is something that jenkins normally adds (I don't involve jenkins in local builds):

```
==> 1-nec-hob.emii.org.au: [2016-07-27T08:38:52+10:00] WARN: Retrying 1/3 '/mnt/source-code/geoserver-build/src/main/target/geoserver-1.0.0-imos.war.md5'
==> 1-nec-hob.emii.org.au: [2016-07-27T08:38:54+10:00] WARN: Retrying 2/3 '/mnt/source-code/geoserver-build/src/main/target/geoserver-1.0.0-imos.war.md5'
==> 1-nec-hob.emii.org.au: [2016-07-27T08:38:56+10:00] WARN: Retrying 3/3 '/mnt/source-code/geoserver-build/src/main/target/geoserver-1.0.0-imos.war.md5'
==> 1-nec-hob.emii.org.au: [2016-07-27T08:38:58+10:00] FATAL: Could not access '/mnt/source-code/geoserver-build/src/main/target/geoserver-1.0.0-imos.war.md5'
=
```

Perhaps return this to not requiring an md5 file for local artifacts?

lwgordonimos:
@jonescc Thanks, this is a bug for local files introduced by https://github.com/aodn/chef/pull/357. It does not consider local files, and obviously needs to!

PR #374, opened 2016-07-27
Fix error introduced when using a local artifact path

lwgordonimos:
@jonescc will resolve https://github.com/aodn/chef/issues/370

The summary of the change is that the Fetcher.download (and related) methods should not be exit points for the Chef run, and should instead return false if they fail, since the calling code uses the "if Fetcher.download then etc.." pattern in order to determine success anyway.


jonescc:
Works well

PR #379, opened 2016-08-10
* update compliance checker remote_file resources to use :create rath…

lwgordonimos:
…er than :create_if_missing

* resolve redirects to get the real file name rather than the HTTP redirect, due to easy_install using the file name to determine version

### DIFF
--- a/cookbooks/imos_po/recipes/netcdf_checker.rb
+++ b/cookbooks/imos_po/recipes/netcdf_checker.rb
@@ -8,13 +8,26 @@
 #
 # Deploys the IOOS netcdf checker
 
-netcdf_checker_filename  = File.basename(node['imos_po']['netcdf_checker']['url'])
+require 'uri'
+require 'net/http'
+
+def follow_redirects(uri)
+  current_uri = URI(uri)
+  r = Net::HTTP.get_response(current_uri)
+  if r.code == "301" or r.code == "302"
+    current_uri = follow_redirects(r.header['location'])
+  end
+  current_uri.to_s
+end
+
+netcdf_checker_real_url = follow_redirects(node['imos_po']['netcdf_checker']['url'])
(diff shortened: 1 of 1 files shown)


cookbooks/imos_po/recipes/netcdf_checker.rb
```diff
@@ -32,17 +45,17 @@
...
-  source   node['imos_po']['netcdf_checker']['cc_plugin_url']
+  source cc_plugin_real_url
   mode     0644
-  action   :create_if_missing
+  action :create
 end
 
 execute "install cc-plugin-imos" do
   user 'root'
   command "easy_install --find-links=#{Chef::Config[:file_cache_path]} --allow-hosts=None --always-unzip #{cc_plugin_full_path}"
```
julian1: I'd like to see the version controlled via the databag. It's not as convenient, but kind of required when using test environments, which is where I've run into this issue. It would also make this consistent with the way we manage almost all other artifacts - wars, talend jobs. 

@mhidas, do you have an opinion on this?

julian1: After speaking with @mhidas and @lwgordonimos, it was suggested that using the Jenkins pipeline would be best. 

A pipeline would keep the artifacts distinct on test and prod - while also supporting easy promotion/deployment, and without the need for a supporting PR to change a databag version number.

mhidas: Just a thought on the Jenkins pipeline idea... would that also allow easy rollback to an earlier version? Or does that only give the option to deploy (or not) the latest code in the git repo?

lwgordonimos: Probably similar in terms of ease of rollback... either way, it definitely makes deployment to prod a more conscious decision, while hopefully making non-prod changes less onerous.

PR #386, opened 2016-08-25
Fix previous incorrect fetcher_jenkins changes

lwgordonimos:


lwgordonimos:
@anguss00 following on from https://github.com/aodn/chef/pull/383... the previous change worked in the scenario we usually use it in but broke unit tests.

PR #391, opened 2016-09-02
Process lock celery worker tasks, prevent restart of running process …

anguss00:
…due to DS SCM change

Retain state of running worker processes for loggin/debugging

### DIFF
--- a/cookbooks/imos_po/recipes/watches.rb
+++ b/cookbooks/imos_po/recipes/watches.rb
@@ -174,16 +174,17 @@ if node['imos_po']['data_services']['watches']
   ruby_block "celery_po_supervisor" do
     block do
       Chef::Recipe::WatchJobs.get_watches(data_services_watch_dir).each do |job_name, watchlist|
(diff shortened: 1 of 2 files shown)


cookbooks/imos_po/recipes/watches.rb
```diff
@@ -174,16 +174,16 @@
...
-        f.command "celery worker --queues=#{job_name} --config=#{celery_config} -A tasks -c #{node['imos_po']['data_services']['celeryd']['max_tasks']}"
+        f.command "celery worker --queues=#{job_name} --config=#{celery_config} --pidfile=#{pidfile}  -A tasks -c #{node['imos_po']['data_services']['celeryd']['max_tasks']}"
         f.directory node['imos_po']['data_services']['celeryd']['dir']
         f.user po_user
         f.run_action :enable
-        f.run_action :restart
+        f.run_action :start
       end
     end
-    subscribes :create, 'git[data_services]',     :delayed
```
lwgordonimos: This subscription should be kept in to make sure it still runs after the repo is updated.

PR #392, opened 2016-09-09
Add support for configuring the wps file expiry time

anguss00:


lwgordonimos:
@anguss00 this will also need to be changed in https://github.com/aodn/chef-private/blob/master/aws/ebprep/conf/templates/aodn-portal/Portal.groovy.template

anguss00:
Good point!

anguss00:
Just putting up the chef-private PR now

julian1:
Why is this in Portal config rather than geoserver config?

anguss00:
@julian1 only because of the email footer content that's sent out after a ggd job is initiated. It's just to configure how long the expiry time is in that email.

PR #394, opened 2016-09-13
prevent supervisord issuing sigkill leading to orphaned celery processes #431

julian1:
https://github.com/aodn/backlog/issues/431

I've tested this, by checking that chef writes the jobs files to /etc/supervisor.d correctly.

And by manually installing supervisord and configuring the following test job. 

After using supervisorctl to stop the job, supervisord sends it a SIGTERM signal, but never follows up with a SIGKILL due to the increase to the stopwaitsecs timeout. 

Instead, the job exits cleanly after terminating.

```
root@systest:/home/jfca# cat mytrap.sh 
#!/bin/bash

function sigterm {
  # Your cleanup code here
  echo 'got SIGTERM'
}

function sigkill {
  # Your cleanup code here
  echo 'got SIGQUIT'
}

trap sigterm SIGTERM
trap sigkill SIGQUIT 


echo "this process id: $$"

while :
 do
   sleep 1
   count=$(expr $count + 1)
   if [ "$count" -gt 60 ]; then
        echo 'terminating gracefully'
        exit 0
   fi
   echo $count
 done
```

### DIFF
--- a/cookbooks/imos_po/recipes/watches.rb
+++ b/cookbooks/imos_po/recipes/watches.rb
@@ -182,6 +182,8 @@ if node['imos_po']['data_services']['watches']
         f = Chef::Resource::SupervisorService.new("celery_po_#{job_name}", run_context)
         f.autostart true
         f.command "celery worker --queues=#{job_name} --config=#{celery_config} --pidfile=#{pidfile}  -A tasks -c #{node['imos_po']['data_services']['celeryd']['max_tasks']}"
+        # ie. one year. to avoid sigkill which will orphan child processes
+        f.stopwaitsecs 31536000
         f.directory node['imos_po']['data_services']['celeryd']['dir']
         f.stdout_logfile ::File.join(supervisor_child_logdir, "#{job_name}-stdout.log")
         f.stdout_logfile_maxbytes node['imos_po']['data_services']['supervisor']['stdout_logfile_maxbytes']


julian1:
It would have been nice if stopwaitsecs could take 0 or -1 to disable the SIGKILL entirely. Unfortunately, it doesn't look like that is supported,

https://github.com/Supervisor/supervisor/blob/master/supervisor/process.py#L422

julian1:
bump

julian1:
I'll add, that I see this as fairly low risk to infrastructure. jobs will still be sent a term signal. We're merely increasing the timeout in a way to prevent supervisord sending a sig-kill, which will leave child processes continuing to run.

anguss00:
Me gusta

julian1:
Chef successfully updated config files. But supervisord has been running since July.  So, I think we will need to restart it - to pick up the change.

```
root@10-aws-syd:/etc/supervisor.d$ ps aux | grep supervi
...
root     21613  0.1  0.0  64688 10792 ?        Ss   Jul07 185:06 /usr/bin/python /usr/local/bin/supervisord -c /etc/supervisord.conf
```

lwgordonimos:
Supervisord itself doesn't need a restart, unless you change /etc/supervisord.conf itself. The child services need a restart to pick up the change to their own config. We should revisit the Chef change to not restart services, because I think this supersedes that change now...

julian1:
I think there's still some uncertainty. 

Since it's not the child processes that will send the TERM or KILL process but the root process. This is slightly different to other job-specific flags that only the child process needs to know about in order to change behavior. 

In addition, supervisord.conf specifically includes the child-processes as a kind of file aggregation kind of operation, which implies that the root process has to have knowledge about some child configuration.

lwgordonimos:
@julian1 as discussed, just to record this...

The supervisor cookbook runs 'supervisorctl update' whenever one of the child services is sent the "enable" action (which is every Chef run). This should cause the entire supervisord config to be reread(including the 'includes'), and so the updated signalling config should work as expected for the queues. Not sure what effect this would have on the global options though, or whether this would still need a restart of supervisord itself.

https://github.com/poise/supervisor/blob/7c9aa287f66327a062287c1cb4b97b225d886e7c/providers/service.rb#L87

Interestingly, the update command might actually be performing a sneaky restart due to "supervisorctl update" stating: "Restarts the applications whose configuration has changed."

julian1:
So the job behaves and can't be orphaned.

But I'm a bit concerned that,

```
supervisorctl stop test
```

will actually hang until the job finishes. 

I didn't expect the supervisorctl to block after firing off the TERM signal at the worker task.

If another instance of supervisorctl is launched, it reports the job is finished - even the process is still running.

So it seems that supervisorctl doesn't actually have any represent for the state between sending the TERM and then waiting for the worker to finish - when there isn't an explicit KILL.

lwgordonimos:
@julian1 See: https://github.com/Supervisor/supervisor/blob/master/supervisor/process.py#L642

As suspected, it will sit in an event loop until the "delay" seconds is 0.

julian1:
Correction, it does have a state for the intermediate condition,

```
root@14-nec-hob:/etc/supervisor.d$ supervisorctl status test
test                             STOPPING
```

julian1:
If chef is idempotent in terms of restarting, it's not quite as bad.

julian1:
Sending explicit SIGTERM outside supervisorctl's control, will restart the job gracefully. 

Two options would be, 
  - cron job to send the sigterm
  - change the chef code to issue a sigterm directly instead of using supervisorctl update start actions

julian1:
Hmmm, in fact sigterm is irrelevant. It's just that a process will be expected to ignore it.

The issue is to have chef update supervisord's config config.

PR #395, opened 2016-09-14
Two stage deployment of schema plugins from war and jenkins build

anguss00:
Rework of schema plugins deployment:
- Deploy skeleton data dir on first creation of geonetwork
- Deploy defined core schema plugins from exploded GN war
- Deploy defined additional plugins from schema plugins pipeline


cookbooks/imos_webapps/definitions/geonetwork.rb
```diff
@@ -79,10 +66,40 @@
...
+      zip_file_cache_path = File.join("#{Chef::Config[:file_cache_path]}", "#{job_name}.gz")
+      remote_file zip_file_cache_path do
+        source download_url.to_s
+      end
+
+      core_schema_plugins_relative_path = File.join("config", "schema_plugins")
+      core_schema_plugins_destination_path = File.join(data_dir, core_schema_plugins_relative_path)
+
+      begin
+        tarball zip_file_cache_path do
```
julian1: Is it not possible to use the regular Jenkins artifact fetcher for this, instead of going down the path of writing specialized code?

That would pull-in the md5 artifact check behavior to avoid doing anything unless the files genuinely differ from what's on the server.

If the artifact checker can't handle tarballs, and can't be adapted to do so - then I would suggest changing the Jenkins schema-plugins jobs to create zips rather than tgz.

The only reason I used tgz for those, is because it's more standard in linux. 

But perhaps if there are other's that will also use these schema-plugins (eg. IMAS) and under a windows-based GN, zip would be better?.

cookbooks/imos_webapps/definitions/geonetwork.rb
```diff
@@ -79,10 +66,40 @@
     end
     only_if {
       ! File.exists?(core_schema_plugins_lock_file) ||
```
julian1: I don't understand the purpose of needing a lock file. Can't we just use the expected location of a deployed plugin, inside the exploded war to whether we need to deploy the plugins?

julian1:
Also, would it make sense to treat all the schema-plugins the same. By defining them in a databag with their full uris.  

That way - then whether the schema-plugin is sourced from Jenkins (most schema-plugins) or from inside the war (the core iso19139) is merely treated as one property of the chef resource.

anguss00:
Change up to enforce dat idempotency

anguss00:
OK, I think this is ready to go now

julian1:
@anguss00 I'm happy to merge, but @lwgordonimos is reviewing.

lwgordonimos:
@anguss00 @julian1 was just running it through my vagrant one more time after the last change. Merging now...

PR #397, opened 2016-09-20
Added sysadmin email cc config

bhasin85:
[Portal PR](https://github.com/aodn/aodn-portal/pull/2331)
[Chef PR](https://github.com/aodn/chef/pull/397)
[Chef-Private PR](https://github.com/aodn/chef-private/pull/2079)


julian1:
Did we want to use bcc?

bhasin85:
@julian1  I think we agreed to go with cc during the planning meeting.

lwgordonimos:
@bhasin85 I think we did decide on CC but I don't believe it was in terms of the sys.admin list... this would put the sys.admin email address in the CC back to the end user, and we don't want users using that as a return address... whatever it is, it should be something that we don't mind end users emailing us on.

bhasin85:
@pblain  What do you think about it?

**Options**

1. cc sys.admin@emii.org.au 
2. bcc sys.admin@emii.org.au
3. cc someOtherEmailAddress@emii.org.au

pblain:
Easiest to switch it to a BCC - there's no real need for the user to see it.

PR #398, opened 2016-09-20
Fix permissions on first creation of GN data dir

anguss00:
I hate having to add another ruby block, but it's the only was I can see to deal with the recursive permissions
Chef directory block doesn't handle recursive permissions properly (bug or feature)

### DIFF
--- a/cookbooks/imos_webapps/definitions/geonetwork.rb
+++ b/cookbooks/imos_webapps/definitions/geonetwork.rb
@@ -96,8 +96,12 @@ define :geonetwork do
     end
   end
 
-  Chef::Log.info("Set ownership of '#{data_dir}' to '#{node['tomcat']['user']}:#{node['tomcat']['user']}'")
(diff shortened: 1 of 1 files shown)


cookbooks/imos_webapps/definitions/geonetwork.rb
```diff
@@ -96,8 +96,12 @@
     end
   end
 
-  Chef::Log.info("Set ownership of '#{data_dir}' to '#{node['tomcat']['user']}:#{node['tomcat']['user']}'")
-  FileUtils.chown_R node['tomcat']['user'], node['tomcat']['user'], data_dir
+  ruby_block "set permissions for #{app_name} data dir #{data_dir}" do
+    block do
+      require 'fileutils'
```
lwgordonimos: Why is this this explicit "require" required in this block, but not the one above, which also uses FileUtils?

PR #399, opened 2016-09-20
send bcc email on error

bhasin85:
[Portal PR](https://github.com/aodn/aodn-portal/pull/2331)
[Chef PR](https://github.com/aodn/chef/pull/399)
[Chef-Private PR](https://github.com/aodn/chef-private/pull/2079)

### DIFF
--- a/cookbooks/imos_webapps/templates/default/gogoduck/gogoduck.groovy.erb
+++ b/cookbooks/imos_webapps/templates/default/gogoduck/gogoduck.groovy.erb
@@ -3,6 +3,7 @@ grails.serverURL = "<%= @custom_parameters['https'] ? 'https' : 'http' %>://<%= 
 // Change default email for IMOS
 grails.mail.default.from = "<%= @custom_parameters['default_email_from_address'] || 'aodnoffice@aodn.org.au' %>"
 grails.mail.default.cc = "<%= @custom_parameters['default_email_cc_address'] || 'sys.admin@emii.org.au' %>"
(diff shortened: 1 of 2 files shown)


cookbooks/imos_webapps/templates/default/gogoduck/gogoduck.groovy.erb
```diff
@@ -2,7 +2,7 @@ grails.serverURL = "<%= @custom_parameters['https'] ? 'https' : 'http' %>://<%=
 
 // Change default email for IMOS
 grails.mail.default.from = "<%= @custom_parameters['default_email_from_address'] || 'aodnoffice@aodn.org.au' %>"
-grails.mail.default.cc = "<%= @custom_parameters['default_email_cc_address'] || 'sys.admin@emii.org.au' %>"
+grails.mail.default.bcc = "<%= @custom_parameters['default_email_cc_address'] || 'sys.admin@emii.org.au' %>"
```
julian1: Why not expose both to chef, so that it can be configured as we want? The cost is only one line of code.

Also, the naming mismatch is a bit odd. The grails var is ```grails.mail.default.bcc``` while the chef attribute name is ```default_email_cc_address```. 

There's a high chance of confusion if cc is specified in the node/databag, but emails are sent bcc. It effectively forces a developer/user to read the code, to know what it does.

bhasin85: I think I forgot to update cc to bcc here (Which created the confusion). I have added the config for both cc and bcc.

PR #405, opened 2016-10-19
Add capability to add Tomcat instance-specific environment variables

lwgordonimos:
@jonescc the tomcat init script looks for this /etc/default/tomcat/$NAME file for inclusion if it exists, so would be interested to see how this works... keeps the variables off the general environment and also the tomcat command line... need to verify that the application can pick them up ok still.

### DIFF
--- a/cookbooks/tomcat/definitions/tomcat_instance.rb
+++ b/cookbooks/tomcat/definitions/tomcat_instance.rb
@@ -17,6 +17,8 @@ define :tomcat_instance do
   logging_properties  = ::File.join(instance_dir, 'conf', 'logging.properties')
   version_sh          = ::File.join(instance_dir, "bin", "version.sh")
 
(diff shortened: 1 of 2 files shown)


lwgordonimos:
Have stared at it for a few more minutes and am not keen on the creds being inline in the node file.

Will revisit with a view to pulling them from a data bag instead...

lwgordonimos:
@jonescc updated so it pulls the AWS creds from a databag alongside any other non-sensitive variables defined directly in the node file (if ever required)

lwgordonimos:
Just for completeness, this hooks into this existing capability in the init script, https://github.com/aodn/chef/blob/master/cookbooks/tomcat/templates/default/tomcat.erb#L92

jonescc:
Looks good - works in vagrant.

PR #421, opened 2016-11-14
SRS SST - use generic timestep harvester

lbesnard:


lbesnard:
merging myself because I realise I've made already 2 other PR with similar mod

PR #426, opened 2016-11-22
Git managed jenkins

bhasin85:
OK, so this is ready for review and POTENTIAL merge.
My two cents is that we should roll up a systest jenkins and test this out before deploying this to prod. It's a lot of changes to the build system.
@lwgordonimos @bhasin85 @julian1 @jonescc


jonescc:
Ran this up from scratch in vagrant and compared to existing jenkins.

Looks to be missing the talend installation directory for building harvesters?

Also missing the tools directory - is this a problem?

cookbooks/imos_jenkins/files/default/.gitignore
```diff
@@ -0,0 +1,48 @@
...
+*.swp
+.owner
+terracotta/
+updates/
+fingerprints/
+config-history/
+talend/
+plugins/
+# Only get the config.xml from any Jobs directory
+jobs/*/builds
```
jonescc: Could you use negated patterns to more concisely specify only config.xml files in the jobs directory?

e.g.

jobs/**
!jobs/**/config.xml

cookbooks/imos_jenkins/files/default/.gitignore
```diff
@@ -0,0 +1,48 @@
...
+jobs/*/modules
+jobs/*/htmlreports
+jobs/*/last*
+jobs/*/next*
+jobs/*/*.csv
+jobs/*/*.txt
+jobs/*/*.log
+jobs/*/workspace
+jobs/*/modules
+# Process jobs that exist in folders
```
jonescc: Do we have any of these?

cookbooks/imos_jenkins/files/default/.gitignore
anguss00: Yep, definitely @jonescc. Sounds like a great idea

jonescc:
The scm plugin appears to be updating the repo automatically with any changes made in/by jenkins - for example - its updating my user file with the last authorisation datae/time - do we need that functionality or could we just update config from the repo directly.

anguss00:
Yeah, that's a good point @jonescc. Will look into that

anguss00:
And will remove the talend dir from the gitignore so we'll get that and tools in the prod repo.

jonescc:
Perhaps installing talend for compilation of talend harvesters should be kept separate from config?  I don't think its considered good practice to include lots of binary files in git.   Dan used to pull down the talend install from binaries.aodn.org.au

jonescc:
The ci-config repo appears to contain a lot of extra things that I wouldn't have thought would be considered config/need to be version controlled, logs caches etc  are we going to go through this and remove all the unnecessary cruft?

cookbooks/imos_jenkins/files/default/.gitignore
jonescc: Actually, wondering why we need to put .gitignore in chef at all - why not leave it in git?

jonescc:
Do we use libraries/view.rb any more?  perhaps delete

cookbooks/imos_jenkins/files/default/.gitignore
anguss00: Agreed. Will remove.

anguss00:
Latest commit addresses Talend and deleting unused files and gitignore @jonescc

jonescc:
@bhasin85 , not quite sure what the intended workflow is for making changes to prod here.  As noted above the scm synch plugin appears to publish any changes made through the gui to the linked repo so when I ran this up in vagrant any changes I made were being pushed to the config master branch.

anguss00:
I would be in favour of having two jobs within Jenkins: one to push the current configuration /home/jenkins up to scm, and another to sync/overwrite the local config with that present in scm. This allows us to push up changes, and reset local config to a known good value if required. I think that covers the workflow requirements

bhasin85:
@anguss00  I think creating 2 jobs will be more clean and will reduce the number of commits, which is too noisy. Also, it will give us the needed workflow as you mentioned. SCM Sync Config Plugin is good if you do not want a workflow.

anguss00:
Ok, production config is here https://github.com/aodn/ci-config

anguss00:
OK, this should be ready for another look @jonescc, the tools dir is not created on a fresh chef run so I believe is old config. Prod config is up at the ci-config repo.

cookbooks/imos_jenkins/attributes/default.rb
```diff
@@ -16,15 +17,33 @@
...
+    'postbuildscript',
+    'job-log-logger-plugin',
+    'maven-plugin',
+    'role-strategy',
+    's3',
+    'ssh-slaves',
+    'throttle-concurrents',
+    'xvfb',
+    'token-macro',
+    'scm-sync-configuration',
```
jonescc: Do we need to install the scm synchronisation plugin if we are not using it?

jonescc:
The current config repo has all jobs disabled.  Do we need to be able to enable for prod disable in jenkins?

cookbooks/imos_jenkins/recipes/managed_master.rb
```diff
@@ -8,20 +8,54 @@
...
+# Overwrite local files with those present in SCM
+execute 'init_jenkins_scm' do
+  command "git rev-parse --is-inside-work-tree || { git init && git remote add origin #{scm_repo} && git fetch && git reset --hard origin/master; }"
+  cwd jenkins_home
+  user node['imos_jenkins']['user']
+  group node['imos_jenkins']['group']
+  notifies :run, "execute[set_git_username]", :immediately
+end
+
+#Set git global config for jenkins scm-sync-configuration plugin
```
jonescc: Perhaps don't need the following either if we aren't using the plugin?

anguss00: Agreed, I'll delete here too

anguss00:
OK, associated PR to remove slave node is here https://github.com/aodn/ci-config/pull/2

anguss00:
I think this beast is ready for more review... even more deletions and less additions @jonescc

lwgordonimos:
letitburn

jonescc:
here

anguss00:
I know what you're thinking punk, did I run six build jobs or five? As a matter of fact, in all the excitement I kinda lost track myself...

julian1:
here

anguss00:
Oh Brick...

PR #427, opened 2016-11-22
Add awscli to Jenkins nodes

lwgordonimos:


bhasin85:
@lwgordonimos  I see some python packages inside node_common.rb.  Do you reckon it is worth moving all of them in one place?

PR #435, opened 2016-12-06
Fix fullstack node to work correctly from external

anguss00:
Validated by NIWA


anguss00:
Paired with @lwgordonimos

PR #442, opened 2016-12-12
Add screen to list of packages in data-services recipe

lwgordonimos:
@mhidas Already installed on 4-nec-hob, but added here for completeness!


mhidas:
Thanks!

PR #446, opened 2016-12-20
Override jenkins global var at runtime

bhasin85:
Related [ci-config PR](https://github.com/aodn/ci-config/pull/7)
This file will be referenced from the Jenkins job which will push the artifact to the S3 bucket. We will need to move the bucket name to a global variable (S3_ARTIFACT_BUCKET) in Jenkins. 

geoserver_edge config jenkins


cookbooks/imos_jenkins/recipes/managed_master.rb
```diff
@@ -51,3 +51,13 @@
   user node['imos_jenkins']['user']
   group node['imos_jenkins']['group']
 end
+
+if Chef::Config['dev']
```
lwgordonimos: As discussed, this should be a template drawing credentials from credentials databags(and any other things we might need in per-environment environment variables). This allows private data from chef-private. Remember chef = code, chef-private = config.

 Also should use chef only_if/not_if rather than Ruby "if".

anguss00:
Looks good from my end... Only thing we need to do is add the env.properties file to the ci-config gitignore

PR #452, opened 2017-01-11
create databags for ghrsst datasets with generic timestep harvester

lbesnard:


jonescc:
Pity, its still quite verbose isn't it.  Looks like programmatic generation is best anyway.

PR #454, opened 2017-01-12
Add workaround for PO box corrupt Python libs

lwgordonimos:
Couple of chef/poise bugs means logic not possible in chef resources, so idempotency handled in the shell command itself. Will be tidied up/removed when better library handling for Python is in place.


anguss00:
Unfortunate, but defo looks like a fix is in the pipeline for this

PR #461, opened 2017-01-18
Pin jenkins plugins to the latest versions for 2.41 upgrade

anguss00:
Related https://github.com/aodn/chef-private/pull/2219


anguss00:
Bumpity

PR #463, opened 2017-01-20
Chef refresh

lwgordonimos:
~~~DO NOT MERGE YET (further testing required to cover all nodes/recipes)~~~

Please review. Should be testable by running up in Vagrant. I have used the chef-refresh and chef-private-refresh branches to Vagrant all nodes successfully(with the exception of 13-aws-syd, which has a known issue in Vagrant: https://github.com/aodn/chef/issues/470)

* cleanups of unused code
* version bumps of cookbooks
* no github cookbooks (only supermarket and local)
* changes to support Ubuntu 16.04 (while retaining support for 12.04)
* changes to support latest Chef (while retaining support for 12.3.0)
* use custom '*.d' config files to extend Squid and APT config rather than overwriting upstream templates


bhasin85:
Related [PR](https://github.com/aodn/chef-private/pull/2228)

anguss00:
So I'm getting a "databags must contain a hash or a mash error" on 8-aws-syd.. bit weird. Not much has changed on there... and the databags are DEFINITELY there

PR #466, opened 2017-02-03
Exit talend-trigger op on error within handling of individual file sl…

anguss00:
…ices
Fix for https://github.com/aodn/content/issues/258. 

If any exception occurs in the file handing stage, it should be caught, raised and the execution bugs out at that point. The current slice AND all successive slices will not be executed and the log is placed in the error dir at that point.

### DIFF
--- a/cookbooks/talend/files/default/talend-trigger.rb
+++ b/cookbooks/talend/files/default/talend-trigger.rb
@@ -15,6 +15,17 @@ end
 @config = nil
 @noop = false
 
+module TriggerOperations
+  class FileTriggerException < RuntimeError
+  end
+
+  class SymbolicLinkException < FileTriggerException
+  end
+
+  class FileProcessingFailureException < FileTriggerException
+  end
+end
+
 def prepare_file_list(objects)
   file_list = Tempfile.new('harvester_file_list')
   objects.each do |object|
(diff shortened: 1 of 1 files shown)


cookbooks/talend/files/default/talend-trigger.rb
```diff
@@ -62,17 +73,15 @@ def prepare_file(tmp_base, file, is_deletion = false)
       target_dir = File.join(tmp_base, File.dirname(index_as))
       target_file = File.join(tmp_base, index_as)
 
-      begin
-        FileUtils.mkdir_p(target_dir)
-        FileUtils.ln_s(real_file, target_file)
-      rescue Exception => e
-        $logger.fatal e.message
-        $logger.fatal "Error in symbolic linking '#{real_file}' => '#{target_file}'. Check to see if '#{target_file}' already exists'"
```
julian1: that's horrible

cookbooks/talend/files/default/talend-trigger.rb
```diff
@@ -62,17 +73,15 @@ def prepare_file(tmp_base, file, is_deletion = false)
...
+      FileUtils.mkdir_p(target_dir)
```
julian1: and, so this is needed because we actually preserve the entire directory structure implied by the key. rather than change / to - .

cookbooks/talend/files/default/talend-trigger.rb
```diff
@@ -148,15 +157,16 @@ def match_and_execute(tmp_base, files)
     $logger.info "--------------------"
     $logger.info files_not_processed
     $logger.info "--------------------"
-    retval += 1
```
julian1: that's horrible too!

cookbooks/talend/files/default/talend-trigger.rb
```diff
@@ -170,11 +180,14 @@ def handle_files(files, delete = false, max_files)
...
-      retval = match_and_execute(tmp_base, files_to_index)
+
+      retval += match_and_execute(tmp_base, files_to_index)
     }
   end
-
   return retval
+rescue => e
+  $logger.fatal e.message
+  return 1
```
julian1: would it be worthwhile adding a comment to indicate that this is the catch all for all exceptions, and then return 1 indicating failure. Just for others to be able to comprehend what's going on?

anguss00:
Made that change @julian1

PR #472, opened 2017-03-15
Remove checksums check when retrieving talend files

akashisama:
By default remote_file will use the If-Modified-Since headers to decide whether to download a new file, which is much better than using checksums since we won't have to remember to update the cookbook when we change the files.

### DIFF
--- a/cookbooks/imos_devel/attributes/talend.rb
+++ b/cookbooks/imos_devel/attributes/talend.rb
@@ -5,13 +5,13 @@ default['imos_devel']['talend']['user'] = "root"
 
 default['imos_devel']['talend']['packages'] = [
     {
-        "source_url" => "http://binary.aodn.org.au/static/talend/TOS_DI-r96646-V5.1.3.zip",
+        "source_url" => "https://s3-ap-southeast-2.amazonaws.com/imos-binary/static/talend/TOS_DI-r96646-V5.1.3.zip",
         "unzip_filter" => "TOS_DI-r96646-V5.1.3/*",
         "install_to" => "",
         "unzip" => true
     },
     {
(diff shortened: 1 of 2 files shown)


lwgordonimos:
@akashisama the checksums also give guarantees that the file is the *expected* file, not just whether it has the file in the cache already (which is almost more of a side effect!).

Might need to determine historically why this was designed in this way to make sure it's not overly loosening the process. What was the driver for this change? Seems like it saves a tiny extra step when updating the file and losing some integrity checking in the process?

akashisama:
According to the documentation at https://docs.chef.io/resource_remote_file.html the checksum option is intended for making sure the file doesn't get redownloaded. "Optional, see use_conditional_get. The SHA-256 checksum of the file. Use to prevent a file from being re-downloaded. When the local file matches the checksum, the chef-client does not download it."

Driver of this change is that @jonescc updated one of the files but didn't realise he had to alter the checksum in the chef recipe.

lwgordonimos:
Ok now worries seems reasonable, looks like it's not really for integrity in that case. Just need to make sure the S3 web server behaves as expected, which it sounds like it does out of the box: http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html#RESTObjectGET-requests :)

akashisama:
Ok going to do some testing to confirm

julian1:
Should probably be factored out of the default attributes and into a databag.

akashisama:
errg I stuffed up that commit, just ignore the removal of the symlinks

akashisama:
Ok I've tested, and it's working well with latest commit. Re placing it in a databag, i look at the databags folder and it's already got so much stuff, is it worth placing an extra databag in it just for this?

PR #473, opened 2017-03-15
Remove nodes environment variable

akashisama:
Pointing this anywhere other than 'nodes' causes the Nagios cookbook to fail because of the node searches it does. So we shouldn't provide the option to change nodes location.


bhasin85:
@akashisama Can we close/merge this PR?

PR #475, opened 2017-03-15
Fix Apache ports.conf to include NameVirtualHost again after upstream…

lwgordonimos:
… change

@anguss00


anguss00:
3 min to spare

PR #485, opened 2017-04-12
A recipe that installs the Geowebcache Seeder

akashisama:
, and a cron job that runs it daily


akashisama:
Not merging for now, @bhasin85 is gonna to try to lambda it instead.

PR #489, opened 2017-04-27
Add systemd cookbook and modify systemd config for Postgres servers t…

lwgordonimos:
…o fix semaphore issue

EDIT: do not merge yet. I would like to look at a more generic systemd cookbook, rather than postgres specific


Berksfile.lock
```diff
@@ -89,6 +89,7 @@ DEPENDENCIES
   sudo
   supervisor
   sysctl
+  systemd
   talend
```
anguss00: Do we need to modify the lock file? Not a biggie, just thought that changes in the Berksfile would automatically pass through to the lock when run

lwgordonimos: As discussed offline, this is just storing the changes generated by the berks command after adding the cookbook.

anguss00: Ah sorry @lwgordonimos, clean forgot!

PR #494, opened 2017-05-09
Overwrite local changes of geoserver config with chaf managed content

anguss00:
As commented, this is a hack. See https://github.com/aodn/issues/issues/44 for clarification on why this is required. 
This is the least nasty way @lwgordonimos and I could think of to trigger an overwrite of local changes while still using the git cookbook. Otherwise, it's a matter of running ALL the git commands locally via execute....


anguss00:
Changes made @lwgordonimos

cookbooks/imos_webapps/definitions/geoserver.rb
```diff
@@ -48,10 +49,39 @@
...
       action     :sync
       notifies   :create, "ruby_block[#{data_dir}_geoserver_injected_variables]", :immediately
+      notifies   :create, "file[#{data_dir}_create_temp]", :immediately
       user       geoserver_data_bag['git_user']   || node['tomcat']['user']
       group      geoserver_data_bag['git_group']  || node['tomcat']['user']
     end
 
+    # HACK: this is to make sure that locally modified files get overwritten by Chef.
+    # Behaviour in git cookbook means that sync action does not do this by default.
+    # The sync action simply compares the SHA of the local HEAD commit with the remote HEAD - y u do dis
```
lwgordonimos: Maybe random file name just to make sure no collisions with the hardcoded name?
e.g.
```ruby
require 'securerandom'
temp_file_name = SecureRandom.uuid

path File.join(data_dir, temp_file_name) # also, no need to format string when it only contains one variable!

command "git add #{temp_file_name}"
```

anguss00:
Added some uuid action @lwgordonimos

bhasin85:
@anguss00 @lwgordonimos Can we close/merge this PR?

lwgordonimos:
Closing, as it's highly debatable whether the original issue was really an issue and don't want to merge old "hacks" when we are moving platforms anyway...

PR #495, opened 2017-05-09
Remove nagios4php

pmbohm:


pmbohm:
Requires https://github.com/aodn/chef-private/pull/2298
This pull request also contains changes to allow Nagios to run on Xenial, therefore testing can be done in Vagrant (8-aws-syd)

PR #503, opened 2017-05-17
Talend job abos_sofs_fl to harvest data form FluxPulse mooring

mhidas:


mhidas:
Included in https://github.com/aodn/chef/pull/604

PR #508, opened 2017-05-23
Removed pnp4nagios from Nagios

pmbohm:


pmbohm:
Requires aodn/chef-private#2298

PR #510, opened 2017-05-24
Simplify running manual basebackup for hot standby node

jonescc:
With these changes, the steps to setup a hot standby database will be:

* run chef to provision the hot standby node -> chef run will abort at the point a basebackup is required.
* run pg_basebackup (only) - the command to use will be included in the chef log above - the password for the replication user will need to be entered
* rerun chef to finish provisioning the hot standby node -> when complete - replication will be in place


cookbooks/imos_postgresql/definitions/imos_postgresql_cluster.rb
```diff
@@ -84,56 +87,71 @@
...
+    content 'The presence of this file indicates that this is a chef managed cluster'
+  end
+
+  if basebackup
+
+    if basebackup['run']
+      # Run basebackup
+      execute "#{name}-basebackup" do
+        command <<-EOS
+            export PGPASSWORD=#{basebackup['password']};
```
lwgordonimos: You can natively set the environment on the execute resource directly, e.g.
```ruby
environment(
      'PGPASSWORD' => basebackup['password']
    )
```

cookbooks/imos_postgresql/definitions/imos_postgresql_cluster.rb
```diff
@@ -84,56 +87,71 @@
...
+          EOS
+        end
+        only_if { Dir["#{data_directory}/*"].empty? }
+      end
+
+      ruby_block "#{name}-basebackup-in-progress" do
+        block do
+          raise 'basebackup in progress'
+        end
+        only_if { `pgrep pg_basebackup` != ''}
```
lwgordonimos: I believe it should use the return code of the process, and can also ditch the backticks with the guard_interpreter, so could just be:

```ruby
guard_interpreter :bash
only_if 'pgrep pg_basebackup'
```

cookbooks/imos_postgresql/definitions/imos_postgresql_cluster.rb
```diff
@@ -84,56 +87,71 @@
...
-        cmd = Mixlib::ShellOut.new(
-          (<<-EOS
-            pg_lsclusters | grep -q -E '^#{version} +#{name} +';
-          EOS
-          ), :user => 'root'
-        )
-        cmd.run_command
-        cmd.exitstatus != 0
-      }
+      command "rm -rf #{data_directory}/*"
```
lwgordonimos: Could probably Ruby-ify this, e.g. (untested)

```ruby
ruby_block "#{name}-create" do
  block do
    require 'fileutils'
    FileUtils.rm_rf Dir["#{data_directory}/*"]
  end
  not_if  { File.exists?(chef_managed_file) }
end
```
since it's probably unavoidable to use shell commands for Postgres, but not for file manipulation!

jonescc:
@lwgordonimos  - changes made/tested

PR #520, opened 2017-06-04
Install postbuildtask plugin by default

anguss00:


anguss00:
If anyone has time for a quick squiz at this one, we can dod the jenkins task if this is complete and be ready to make up a release plan.

PR #528, opened 2017-06-22
Manual basebackup

lwgordonimos:
Resubmitting PR from this branch after adding fix for SSH key/user creation ordering issue


lwgordonimos:
@pmbohm this fixes the issue you had with bringing up the box from scratch, so should be pretty right to go.

PR #532, opened 2017-06-26
Load Postgres creds from databag to enable different environments to …

lwgordonimos:
…use different creds

To go with https://github.com/aodn/chef-private/pull/2344


cookbooks/imos_postgresql/attributes/default.rb
```diff
@@ -7,6 +7,7 @@
 default[:imos_postgresql][:postgresql_version] = "9.6"
 default[:imos_postgresql][:postgis_version] = "2.3"
 default[:imos_postgresql][:postgresql_service_user] = "postgres"
+default[:imos_postgresql][:postgresql_service_user_databag] = "postgres_systest"
```
anguss00: See comments on chef-private PR

lwgordonimos: You're correct, it should have been postgres_rc, not systest.

PR #536, opened 2017-06-28
CARS - handle monthly split CARS product

lbesnard:


private-sample/data_bags/talend/cars_monthly.json
```diff
@@ -10,7 +10,7 @@
     },
     "event": {
         "regex": [
-            "^CSIRO/Climatology/CARS/.*\\monthly.nc$"
+            "^CSIRO/Climatology/CARS/.*\\monthly.*\\.nc$"
```
ggalibert: We need something like:
```json
"^CSIRO/Climatology/CARS/2009/AODN-product/CARS2009_World_monthly_[0-9]{2}.nc$"
```

PR #538, opened 2017-07-11
Talend job aodn_nsw_oeh to harvest single-beam and multi-beam data

mhidas:
See https://github.com/aodn/harvesters/pull/529


mhidas:
Not quite ready yet... also need to update the metadata parameters.

PR #540, opened 2017-08-02
Geonetwork config manager improvements

anguss00:
Ready for review. Let me know if there's anything I've forgotten to add here

@bhasin85 if you pull down this branch it should be good to go for testing in beanstalk


cookbooks/imos_webapps/files/default/geonetwork/geonetwork-config-manager.rb
```diff
@@ -2,8 +2,7 @@
```
bhasin85: Script fails silently when supplied with the wrong password. 

`sudo /opt/chef/embedded/bin/ruby geonetwork-config-manager.rb -g http://172.28.128.3:8081/geonetwork -c geonetwork-config-geonetwork.json -u admin -p admin1111
`

bhasin85: Do you it will be a good idea to add some logging to know what is happening within the script. Something like 

```
Starting script
Creating user - admin
User created successfully - admin
Adding logo - AAD.gif
Unable to add logo - AAD.gif
Ending script with error
```

anguss00:
Logging added @bhasin85, there are a couple of issues where logo and vocab uploads are attempted where the files are already present. Looking at locking down these so the logs aren't clotted with errors and excess requests, but the functionality works at this stage.

anguss00:
Changes up @bhasin85

bhasin85:
Nailed it! :1st_place_medal: 
giphy

PR #546, opened 2017-08-23
Install aodncore/aodndata from local repo rather than checkout in vag…

lwgordonimos:
…rant


anguss00:
Should we try and cherry-pick this over onto a PR against the new branch?

PR #552, opened 2017-08-29
Various fixes required for po's to use geonetwork-config-manager script

jonescc:
o install dependencies using gem_package instead of chef_gem as system users need to run this script not chef
o append srv/eng to geonetwork url's if required
o upload logo's using http rather than copying to internal upload directory
o use config file in script directory by default so that it doesn't need to be specified when not in the script directory


cookbooks/imos_webapps/files/default/geonetwork/geonetwork-config-manager.rb
```diff
@@ -519,10 +531,10 @@ def main(url, temp_dir, username, password, do_reindex)
   Examples:
     
     Default Chef managed geonetwork configuration:
-    sudo ./geonetwork-config-manager.rb -g "http://localhost:8080/geonetwork" -t "/mnt/ebs/tomcat7/geonetwork/webapps/geonetwork/data/tmp"
+    sudo ./geonetwork-config-manager.rb -g "http://localhost:8080/geonetwork"
     
     Geonetwork configuration with optional parameters
-    sudo ./geonetwork-config-manager.rb -g "http://localhost:8080/geonetwork" -t "/mnt/ebs/tomcat7/geonetwork_portal/webapps/geonetwork/data/tmp" -c "geonetwork-config-geonetwork.json" -u "admin" -p "admin"
+    sudo ./geonetwork-config-manager.rb -g "http://localhost:8080/geonetwork" -c "geonetwork-config-geonetwork.json" -u "admin" -p "admin"
```
anguss00: If we're not putting files in the default temp dir for geonetwork, I think we can drop the sudo command now

jonescc: Yep.  We should drop that.   I'll make appropriate  changes.

jonescc:
Changes made @anguss00

PR #557, opened 2017-08-30
Add error_dir to the po box default pipeline config file

anguss00:
We also need to set up the error directory structure for pipeline_2, and think about how it will coexist with the existing pipelines. But this should fix the issue https://github.com/aodn/zzz-aodn-pipeline-poc/issues/65 immediately


cookbooks/imos_po/files/default/pipeline.conf
```diff
@@ -2,7 +2,8 @@
   "global": {
     "admin_recipients": "email:leigh.gordon@utas.edu.au,email:angus.scheibner@utas.edu.au",
     "upload_uri": "file:///tmp/YOLO",
-    "wip_dir": "/mnt/ebs/wip"
+    "wip_dir": "/mnt/ebs/wip",
+    "error_dir": "/vagrant/src/error"
```
mhidas: Could we make it `/mnt/ebs/error` so it's the same as on RC/prod?

anguss00: Yep, sure thing

anguss00:
done @mhidas

PR #572, opened 2017-09-06
Legacy and new pipeline running simultanously

anguss00:
WIP: not to be merged yet.

@lbesnard @mhidas @lwgordonimos this branch can be used to test pipeline 2. I would suggest moving towards this branch for testing as this will hopefully get merged into master in the near future


bin/vagrant-wrapper.sh
```diff
@@ -177,7 +177,7 @@ po_pipeline_2_box_main() {
     declare -r DATA_SERVICES_GIT_REPO=git@github.com:aodn/data-services
 
     export VAGRANT_STATIC_IP=10.11.12.13
-    export VAGRANT_MEMORY=3072
+    export VAGRANT_MEMORY=8192
```
mhidas: :+1: I've been changing this locally every time I run it!
I also give it 2 CPUs - not sure if that would be worth doing by default, or maybe not really necessary?

anguss00: Yes, happy to give it more. I was testing with 4 actually, made it snappier

cookbooks/imos_po/files/default/pipeline.conf
```diff
@@ -2,7 +2,7 @@
   "global": {
     "admin_recipients": "email:leigh.gordon@utas.edu.au,email:angus.scheibner@utas.edu.au",
     "upload_uri": "file:///s3/imos-data",
-    "wip_dir": "/mnt/ebs/wip",
+    "wip_dir": "/tmp/wip",
```
mhidas: Would it make sense to go the other way, i.e. set both wip_dir and $WIP_DIR to "/mnt/ebs/wip", same as in prod?

lwgordonimos: This file will be a template so the values will be plugged in per environment by chef.

/mnt/ebs doesn't really mean anything for the PO box, and the values are currently provided in the node file, i.e

```json
    "imos_po": {
        "data_services": {
            "create_watched_directories": true,
            "user": "vagrant",
            "group": "vagrant",
            "clone_repository": false,
            "dir": "/vagrant/src/data-services",
            "log_dir": "/vagrant/src/data-services/log",
            "public_dir": "/mnt/public",
            "archive_dir": "/mnt/archive",
            "tmp_dir":     "/vagrant/src/data-services/tmp",
            "incoming_dir": "/var/incoming",
            "error_dir": "/vagrant/src/error",
            "data_dir": "/s3/imos-data",
            "graveyard_dir": "/vagrant/src/graveyard",
            "wip_dir": "/tmp/wip"
        },
```

It shouldn't really matter what the value is, as long as it is reflected in the variables. In fact, I'm in favour of *not* using /mnt/ebs, since it is neither a mount point nor an EBS volume!

Several others use the "/vagrant/src" mountpoint to persist the directory and share it with the host, but not sure how much sense this makes for WIP in particular!

In any case, once templated properly, the file will line up with the current PO box. We can change it if need be, but /mnt/ebs probably isn't correct going forward.

anguss00:
@lwgordonimos does that last commit build in vagrant? I get errors about no distribution for aodncore when trying to install aodndata requirements

mhidas:
I just tried too. It provisions without complaint, but the celery processes don't start up. I'm getting
```
ImportError: No module named _tkinter, please install the python-tk package
```

lwgordonimos:
@mhidas confirmed this is due to matplotlib version 2 requiring 'tkinter' to be installed. The current data-services requirements pins the version to 1.5.1, presumably for this reason.

As you mentioned, you can override the backend, which would probably be the solution if we wanted to run version 2, but is a bit more mucking around when importing:
https://stackoverflow.com/questions/4931376/generating-matplotlib-graphs-without-a-running-x-server

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -175,7 +175,17 @@
...
-pipeline_watches.each do |watch|
+legacy_pipelines = Chef::Recipe::WatchJobs.get_watches(
+    File.join(node['imos_po']['data_services']['dir'], 'watch.d')).keys
+duplicates = pipeline_watches & legacy_pipelines
+
+duplicates.each do |watch|
+  Chef::Log.warn("Legacy pipeline already configured for watch #{watch}. Skipping supervisor service creation.")
+end
+
+watches_to_deploy = pipeline_watches - duplicates
```
mhidas: So in case there is a duplicate watch defined (old & new pipeline), this will leave the old one in place and not deploy the new one, right?

anguss00: Correct. The behaviour I'm working on at the moment is:
If new pipeline defined and old removed from DS watch.d dir, disable old supervisor service and delete old config file.

anguss00:
Fixed the issue with creation of the processing dir @lwgordonimos @mhidas

cookbooks/imos_po/recipes/data_services.rb
```diff
@@ -109,6 +109,7 @@ def imos_po_credentials
   'ERROR_DIR'             => node['imos_po']['data_services']['error_dir'],
   'ARCHIVE_IMOS_DIR'      => node['imos_po']['data_services']['archive_dir'],
   'WIP_DIR'               => node['imos_po']['data_services']['wip_dir'],
+  'PROCESSING_DIR'        => node['imos_po']['data_services']['wip_dir'],
```
mhidas: Should that be`node['imos_po']['data_services']['processing_dir']` ?

PR #579, opened 2017-09-13
Add geonetwork config manager & vocabs in po_pipeline_2 box

mhidas:
Copied from #556 
Works fine for me when running up from scratch, following instructions [here](https://github.com/aodn/chef/blob/master/doc/README.po-box.md#importing-vocabulariesharvesting-geonetwork-records)


mhidas:
BTW, it doesn't seem to be a problem for chef, but should we update private-sample/nodes/po_pipeline_2.json to use 4-space indents like all the other json files? (it currently has 2-space indents)

anguss00:
Great stuff, I can fix that if you like?

mhidas:
Yes please, if you think it's worth the trouble.

PR #580, opened 2017-09-14
Fixes #550 - run harvesters once in configured order

### LINKED ISSUE #550: Geonetwork config manager runs non IMOS Catalogue harvesters twice
### To reproduce 

Run geonetwork config manager with harvesters configured

### What happens

Any harvester not named 'IMOS Catalogue' is run twice

### What II Expect to Happen

Harvetsers are only run once

jonescc:


anguss00:
Looks good @jonescc. This will need a corresponding chef-private change to ensure correct ordering?

jonescc:
Sure.   But I don't think this PR has to wait for that.  It's always run them in the current order with a delay in between them albeit all harvesters except catalogue-imos twice.

PR #582, opened 2017-09-14
Old and new pipelines in PO Box to upload to the same directory

mhidas:
Not sure if this might break anything else, but it works in my box.

Also has the benefit of not losing all the "uploaded" files every time the box is restarted.


anguss00:
Little risk in this PR. Nice one

PR #592, opened 2017-09-27
Give example of how to source files from s3 in po box thredds config

jonescc:


jonescc:
bump..   comments only - should be harmless...

PR #605, opened 2017-10-27
Create local fake s3 bucket by default on 6-nec-hob

anguss00:
https://github.com/aodn/zzz-aodn-pipeline-poc/issues/135


lbesnard:
I thought we decided to create a real s3 bucket ?

anguss00:
Yes we have, I've got to wrap this in a dev mode statement so it doesn't get created in prod

anguss00:
We will definitely create the real bucket

PR #612, opened 2017-11-06
Pl2 fix watches

lwgordonimos:
This should fix the issue with the watches.conf not being updated!


mhidas:
Thanks @lwgordonimos . That does the trick!

PR #617, opened 2017-11-14
Update JQ version

bhasin85:
https://github.com/aodn/aws-wps/pull/140
https://github.com/aodn/ci-config/pull/109
https://github.com/aodn/chef/pull/618


cookbooks/imos_core/attributes/jq.rb
```diff
@@ -1,2 +1,2 @@
-default['imos_core']['jq']['url'] = "http://ap-southeast-2.ec2.archive.ubuntu.com/ubuntu/pool/universe/j/jq/jq_1.2-8~ubuntu12.04.1_amd64.deb"
+default['imos_core']['jq']['url'] = "http://archive.ubuntu.com/ubuntu/pool/universe/j/jq/jq_1.5+dfsg-1_amd64.deb"
```
lwgordonimos: This really should just be a regular apt install... the only thing I can think of was than on Ubuntu Precise, it needed a newer version than in the repos... on Xenial this is just a roundabout way to install the same version but bypassing apt for no good reason.

I'd get rid of the jq recipe completely and just add it here instead:

https://github.com/aodn/chef/blob/master/cookbooks/imos_jenkins/attributes/default.rb#L32

anguss00: Yeah should be pretty low risk TBH

bhasin85: New PR https://github.com/aodn/chef/pull/618

PR #627, opened 2017-11-24
Install compliance checker and cc plugin on jenkins via wheels instea…

anguss00:
…d of easy_install.

We'll need to build and promote the prod artifact of compliance checker before this can be run. Not sure what the current status of that is RE testing/release


lwgordonimos:
As discussed, we need to have a chat with @mhidas before proceeding and promoting the Jenkins artifact, due to the additional commit.

PR #632, opened 2017-11-30
Guard against overwriting pipeline watch dirs if they already exist

anguss00:


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -202,12 +202,18 @@
...
         relative_path = Pathname.new(path)
-        Chef::Log.info("Creating Pipeline 2 watch directory #{path}")
-        wd = Chef::Resource::Directory.new(File.join(node['imos_po']['data_services']['incoming_dir'], relative_path), run_context)
-        wd.user po_user
-        wd.group po_group
-        wd.recursive true
-        wd.run_action :create
+        directory_to_create = File.join(node['imos_po']['data_services']['incoming_dir'], relative_path)
+
+        if not ::File.directory?(directory_to_create)
```
lwgordonimos: This should probably just check whether it exists, otherwise if for some reason there is a regular file with the same name, this would hit the else block and error.

anguss00:
Fixed @lwgordonimos

PR #636, opened 2017-12-05
Refactor out core python packages into new recipe rather than bundlin…

anguss00:
…g into s3config
Fixes https://github.com/aodn/zzz-aodn-pipeline-poc/issues/142


anguss00:
Also double quote action @lwgordonimos 
trololol

PR #639, opened 2017-12-05
Pipeline 2 - add FAIMMS to pobox

lbesnard:


private-sample/data_bags/imos_po_watches/FAIMMS.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "FAIMMS",
+  "path": [ "FAIMMS" ],
+  "handler": "FAIMMSHandler",
```
anguss00: This handler is listed as FaimmsHandler in aodndata, so just needs to be renamed here

lbesnard:
done
 and this one
https://github.com/aodn/chef-private/pull/2612

PR #644, opened 2017-12-07
Install s3cmd and packages as part of separate recipe in core

anguss00:
Together with https://github.com/aodn/chef-private/pull/2609


cookbooks/imos_backup/recipes/default.rb
```diff
@@ -113,6 +113,13 @@
     mock_credentials = false
   end
 
+  # AS: make sure we're using the python s3cmd utility
```
lwgordonimos: Not sure we need a comment here? Installing the package is pretty readable and self explanatory. Coming back in 6 months time to comments like this is more confusing that helpful!

cookbooks/imos_po/recipes/s3_credentials.rb
```diff
@@ -1,6 +1,6 @@
 #
 # Cookbook Name:: imos_po
-# Recipe:: s3cmd
+# Recipe:: s3_credentials
```
lwgordonimos: "s3_credentials" is a bit ambigious. This is specific to s3cmd, so we should be careful not to imply that it is related to any other official S3 config (e.g. boto/awscli). So I'd maybe go for "s3cmd_credentials"?

anguss00:
Changes up @lwgordonimos, also in private

cookbooks/imos_core/recipes/s3_cmd.rb
```diff
@@ -0,0 +1,17 @@
+#
+# Cookbook Name:: imos_core
+# Recipe:: s3_cmd
```
lwgordonimos: One last final thing... can we call this "s3cmd" so it matches the name of the utility, i.e. without the underscore?

anguss00:
Doneskis @lwgordonimos

PR #650, opened 2017-12-20
Manage .aws/config on Jenkins

lwgordonimos:


cookbooks/imos_jenkins/templates/default/aws_config.erb
```diff
@@ -0,0 +1,3 @@
+<% @lines.each do |l| %>
+  <%= l %>
```
anguss00: This extra space here is being translated to the template... do we need to have that syntactically?

lwgordonimos:
@anguss00 the template space was unintentional, but I've changed it a bit more anyway, to be a bit cleaner

anguss00:
Sleek

PR #655, opened 2018-01-30
Delete notify success and error lists on non-prod nodes

anguss00:


anguss00:
Bumpity @lwgordonimos. We'd better merge this one

PR #662, opened 2018-02-21
Upgrade compliance checker core to 3.1.1 from pypi

anguss00:
Together with https://github.com/aodn/ci-config/pull/152 and https://github.com/aodn/python-aodncore/pull/43


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -204,7 +203,7 @@
     cwd node['imos_po']['pipeline']['local_aodndata_repo']
     virtualenv environment_dir
   end
-else
+ else
```
mhidas: Extra space unintended?

mhidas:
Does this apply to v1 pipelines as well? It should be only for v2.

anguss00:
Yes, this will only install compliance checker 3.1.1 in the venv for v2 pipeline

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -171,10 +171,10 @@
   virtualenv environment_dir
 end
 
-cc_artifact_path = Chef::Recipe::PipelineUtils.get_deployed_artifact_path(node['imos_po']['pipeline']['compliance_checker_artifact'], node)
-python_execute 'install_compliance-checker' do
-  command "-m pip install #{cc_artifact_path}"
+python_package 'compliance-checker' do
+  version node['imos_po']['pipeline']['compliance_checker_version']
   virtualenv environment_dir
+  action :upgrade
```
lwgordonimos: My only comment is around this bit. It looks correct, but my level of confidence in the poise behaviour is not overly high. Wondering how we can test this in all the relevant scenarios?

e.g.
* no version installed
* lower version installed
* same version installed
* higher version installed

anguss00: Yep, definitely. I'll test this out with all of those cases.

anguss00: happy

anguss00:
Ready to go @mhidas @lwgordonimos

PR #664, opened 2018-02-28
Remove 'local' declaration

ccmoloney:
'local' declaration of variable seems to be causing failure during run of backup scripts.


lwgordonimos:
That'll teach you for trying to scope variables appropriately.

Also, since we're in Ruby land here in the template, it would be a lot cleaner to replace this:

```bash
 monthly_backup_path=`echo $s3path|sed 's#\/backups#\/backups\/monthly#g'`
```

with a Ruby equivalent. Since all the information is known at the time of templating makes sense to do the work at that point, instead of generating shell commands to execute at runtime.

cookbooks/imos_backup/templates/default/model_pgsql.sh.erb
```diff
@@ -40,7 +40,7 @@ store() {
     # ie: in a 'monthly' folder.
 
     #  Work out the monthly backup path
-    local monthly_backup_path=`echo $s3path|sed 's#\/backups#\/backups\/monthly#g'`
+    monthly_backup_path=`echo $s3path|sed 's#\/backups#\/backups\/monthly#g'`
```
lwgordonimos: As mentioned in other comment, this should be a Ruby variable declared above, and work like s3path in terms of being templated in.

PR #666, opened 2018-03-02
S3 fetcher and pipeline node file

akashisama:


akashisama:
@lwgordonimos I've removed the stack json. I think it should be good to merge.

PR #672, opened 2018-03-23
Script to munge aodn-secrets databags and nodes

anguss00:


bin/munge_secrets.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+rsync -avzh --exclude '.*' ../../aodn-secrets/ ../../chef-private
```
lwgordonimos: Just one thing. Since the current working directory will be the root of this repo, and the other repos checked out here, I would have thought the command should be

```bash
rsync -avzh --exclude '.*' ./secrets/ ./private
```
if the repos are in the expected location?

anguss00: Yup makes sense. I was thinking on the basis of having chef, chef-private and secrets in the same directory level. But if secrets and private are checked out into chef directory then we'll have to make the change.

anguss00: Only thing is, will have to be ``../aodn-secrets/ ../chef-private`` as it is in the /bin dir

anguss00:
Modded @lwgordonimos

PR #673, opened 2018-03-26
Remove imos_dns, imos_firewall and consolidate into imos_core

anguss00:


cookbooks/imos_core/recipes/etc_hosts.rb
```diff
@@ -8,7 +8,7 @@
 #
 
 # A quick recipe to configure /etc/hosts
-node['imos_dns']['etc_hosts'].each do |ip_addr, names|
+node['etc_hosts'].each do |ip_addr, names|
```
lwgordonimos: I'd put this under a top level "imos_core" node so it matches the cookbook structure exactly

anguss00:
@lwgordonimos all done!

PR #674, opened 2018-03-27
Geowebcache tile check

gsatimos:
There is another PR associated with this in the chef-private repo. These need to be tested together in nagios in vagrant.
https://github.com/aodn/backlog/issues/701


cookbooks/nagios/templates/default/adhoc_services.cfg.erb
```diff
@@ -0,0 +1,30 @@
...
+            servicegroups = service['servicegroups']
+          end
+        end  
+       
+%>
+define service {
+  host_name <%= serverless_host['id'] %>
+  service_description adhoc-service-check
+  servicegroups <%= servicegroups %>
+  use default-service
```
jonescc: Should we be picking this up from the service definition as well?

gsatimos:
closed because checking s3 tiles takes about 4-6 mins which is too long for a nagios check to be running. I'm leaving this branch here for the time being because this might get revisited depending on what other options for tile checking work out

PR #675, opened 2018-03-27
Remove depth service perl script - after portal testing

pmbohm:


pmbohm:
For aodn/backlog#748
Also see https://github.com/aodn/chef-private/pull/2761

PR #677, opened 2018-03-28
Prevent Vagrant Nagios from attempting connections to private subnets

lwgordonimos:


lwgordonimos:
@gsatimos this resoles the issue ITS was having with spamming the wireless network from Vagrant Nagios

PR #681, opened 2018-04-17
Source pip 9.0.3 for system and venv python

anguss00:


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -169,10 +169,19 @@
 
 apt_package node['imos_po']['pipeline']['packages']
 
+# Temporary fix for https://github.com/poise/poise-python/issues/107
+node.default['poise-python']['options']['get_pip_url'] = node['imos_po']['pipeline']['pip_download_url']
```
lwgordonimos: Can't you just set default['poise-python']['options']['get_pip_url'] directly in the attributes, rather than defining an imos_po attribute only to assign here?

https://blog.chef.io/2013/12/03/doing-wrapper-cookbooks-right/

anguss00: I wanted to isolate the default node setting to the recipe itself as we're going to get rid of it when poise fix the bug. But I see the point

anguss00:
Done @lwgordonimos

anguss00:
@lbesnard this PR should fix your PO box issues

PR #687, opened 2018-05-07
Fix exception handling in manager script as result is null

anguss00:


lwgordonimos:
Closing as this is being replaced by playbook

PR #688, opened 2018-05-16
Pin netcdf4 to version 1.3.1

anguss00:
Workaround for https://github.com/ioos/compliance-checker/issues/579


lwgordonimos:
Sorry, I was looking into this but ran out of time. This will do for now, but I think the root cause is this other workaround being too aggressive in updating dependencies: https://github.com/aodn/chef/blob/master/cookbooks/imos_po/recipes/pipeline.rb#L188

This hack will do for now, and we can look at whether we can revisit that resource above to make it behave. Maybe the latest poise fixed the issue we were having with python_package at the time?

anguss00:
Yeah fair call @lwgordonimos that --upgrade flag is pretty aggro. I have a feeling that poise is doing things correctly now, as I was able to use the package resource here to downgrade the netcdf4 version

anguss00:
I quickly dashed this through as it was affecting prod

lwgordonimos:
No probs, I don't have a problem with it as an urgent fix ;)

The behaviour we want is "manage the target package according to version requirement, only update dependencies if the target package updates requirements". That's the only way to prevent untested changes flying out of PyPI into production!

anguss00:
Yesss it's a timebomb waiting to explode at the mo

mhidas:
:+1: Thanks guys.

PR #690, opened 2018-05-23
Revert back to using python_execute for local wheels

anguss00:


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -224,15 +224,17 @@
...
+    command "-m pip install --upgrade -c #{constraints_file} --no-deps --force-reinstall -e ."
   end
 
   aodndata_artifact_path = Chef::Recipe::PipelineUtils.get_deployed_artifact_path(node['imos_po']['pipeline']['aodndata_artifact'], node)
-  python_package aodndata_artifact_path do
+  python_execute aodndata_artifact_path do
     virtualenv environment_dir
-    install_options "-c #{constraints_file}"
+    cwd node['imos_po']['pipeline']['local_aodndata_repo']
+    command "-m pip install --upgrade -c #{constraints_file} --no-deps --force-reinstall -e ."
```
lwgordonimos: These are the arguments for an "editable" install, not a wheel install.

It should be something more like:
```
command "-m pip install -c #{constraints_file} #{aodndata_artifact_path}"
```

PR #700, opened 2018-06-06
Cookbook to set up command post box for creating po stacks

anguss00:


cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+# Recipe:: default
+#
+# Copyright 2018, IMOS
+#
+# All rights reserved - Do Not Redistribute
+#
+
+include_recipe 'imos_core::git_deploy_key'
+include_recipe 'imos_devel::chef_dk'
+cloud_deploy_dir = '/opt/cloud-deploy'
```
lwgordonimos: Hardcoded value, should be an attribute

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+
+apt_package 'ruby-bundler'
+
+python_virtualenv '/opt/ansible-virtualenv'
```
lwgordonimos: Hardcoded value, should be an attribute

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+
+git cloud_deploy_dir do
+  repository 'git@github.com:aodn/cloud-deploy.git'
```
lwgordonimos: Hardcoded value, should be an attribute

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+  revision 'master'
```
lwgordonimos: Hardcoded value, should be an attribute

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+  notifies :run, 'execute[install_ruby_deps]', :immediately
+end
+
+execute 'install_ruby_deps' do
+  command 'bundle install --path bundle'
+  cwd "#{cloud_deploy_dir}/tools/ruby"
+end
+
+git '/opt/chef-private' do
+  repository 'git@github.com:aodn/chef-private.git'
```
lwgordonimos: Hardcoded value, should be an attribute

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+  revision 'master'
```
lwgordonimos: Hardcoded value, should be an attribute

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+  ssh_wrapper node['git_ssh_wrapper']
+  action :sync
+end
+
+template '/opt/stack_secrets.py' do
```
lwgordonimos: I think these *static* scripts should probably be moved out to cloud-deploy. Chef should be responsible for setting enviroment variables for them to use, but they aren't really templates

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+  cookbook 'external_templates'
+  source 'command_post/stack_secrets.py.erb'
+  owner 'root'
+  group 'root'
+  mode '0755'
+end
+
+template '/etc/profile.d/create_stack_env.sh' do
```
lwgordonimos: This could remain a template, but it should have dynamic values passed into it, i.e. the attributes I mentioned above ;)

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+  cookbook 'external_templates'
+  source 'command_post/create_stack_env.sh.erb'
+  owner 'root'
+  group 'root'
+  mode '0644'
+end
+
+node['etc']['passwd'].each do |user, data|
```
lwgordonimos: This file could be templated into one place and then we'd just need to set  AWS_CONFIG_FILE in the profile script rather than templating multiple times

https://docs.aws.amazon.com/cli/latest/userguide/cli-config-files.html

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,92 @@
...
+    template "#{aws_dir}/config" do
+      cookbook 'external_templates'
+      source 'aws/dot_aws_config.erb'
+      owner user
+      group data['gid']
+      mode '0600'
+      variables({:config => node['aws_config']})
+    end
+
+    chef_dir = "#{data['dir']}/chef"
```
lwgordonimos: This clone stuff would probably be better in a script, which is called "on demand" to spin up the stack.

Users can't run chef to update it themselves, so we end up with a half chef managed/half manual monstrosity. Better to have it in a script I think that can operate per-user.

anguss00:
Ready @lwgordonimos

cookbooks/imos_command_post/recipes/default.rb
```diff
@@ -0,0 +1,84 @@
...
+
+stackman_root = node[:imos_command_post][:stackman_root]
+
+config_dir = File.join(stackman_root, 'etc')
+virtualenv_dir = File.join(stackman_root, 'python-virtualenv')
+repos_dir = File.join(stackman_root, 'repos')
+
+cloud_deploy_dir = File.join(repos_dir, 'cloud-deploy')
+private_dir = File.join(repos_dir, 'private')
+ebprep_conf_dir = File.join(private_dir, '/aws/ebprep/conf')
```
anguss00: Really small thing, maybe go ```File.join(private_dir, 'aws', 'ebprep', 'conf')```?

PR #703, opened 2018-06-13
Changes required to get rid of aws_jenkins databag

lwgordonimos:


cookbooks/imos_jenkins/recipes/managed_master.rb
```diff
@@ -101,20 +101,10 @@
...
-else
-  credentials_databag_name = node['imos_jenkins']['s3']['credentials_databag']
-end
-
-credentials_databag = Chef::EncryptedDataBagItem.load("passwords", credentials_databag_name)
-
-envvars[:AWS_ACCESS_KEY] = credentials_databag['access_key_id']
-envvars[:AWS_SECRET_KEY] = credentials_databag['secret_access_key']
-envvars[:S3_ARTIFACT_BUCKET] = credentials_databag['artifact_bucket']
+envvars[:S3_ARTIFACT_BUCKET] = node['imos_jenkins']['s3']['artifact_bucket']
```
akashisama: This will set the env variable to point to the production bucket when 13-aws-syd is run up on vagrant. Should put it in an if else statement. By  default a vagrant user shouldn't have access to write to the bucket anyway, but better safe than sorry.

PR #714, opened 2018-07-12
Increase maxImageWidth/Height to default values 2048x2048 in thredds …

ggalibert:
…config.


ggalibert:
https://github.com/aodn/issues/issues/241#issuecomment-403733354

PR #717, opened 2018-08-07
PO node - update pipeline2 imos_po_watches json to match chef-private…

lbesnard:
… ones


lbesnard:
It was not possible to provision the po box and use the pipeline 2 prior to this commit. All the json files still had a old version breaking the pipeline 2. 
The json files were simply taken from the chef-private repo

lwgordonimos:
@lbesnard can you please remove the notify_lists with real email addresses in them and overwrite this commit? Aside from that, should be fine...

PR #723, opened 2018-08-17
Use TLS 1.2 by default for Java applications

anguss00:
Part of the fix for https://github.com/aodn/issues/issues/258


cookbooks/tomcat/attributes/default.rb
```diff
@@ -42,5 +42,5 @@
 
 default['tomcat']['java_options'] = "-Xmx2G -Xms1G -XX:PermSize=128m -XX:MaxPermSize=256m -XX:+HeapDumpOnOutOfMemoryError"
 default['tomcat']['jmx_remote_port_prefix'] = '2'
-default['tomcat']['jmx_options'] = "-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=%d -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false"
+default['tomcat']['jmx_options'] = "-Dhttps.protocols=TLSv1.2 -Djdk.tls.client.protocol=TLSv1.2 -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=%d -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false"
```
jonescc: Don't think this will work - we don't use jmx options when this is deployed

jonescc: Don't always/shouldn't always use jmx options I should say.  Currently have it switched on for geoserver_wps only?

lwgordonimos: https://github.com/aodn/chef/blob/master/cookbooks/tomcat/definitions/tomcat_instance.rb#L26

Might be better to handle in here to enforce it, or just make sure it's added to each relevant node, since they pretty much all have custom java_options anyway.

jonescc: I think in the cookbook is better (mandatory options added to all)

anguss00: Fair call, those options don't really belong in jmx. I'll close and add the java options on a node by node basis.

lwgordonimos: This cookbook is Java version agnostic though, so when upgrading Java separately from this cookbook would then need this changed again, and couples the two separate cookbooks together in an environment specific way.

You could maybe add a *node['tomcat']['additional_java_options']* to the cookbook (which is added by default, but able to be overridden by node files), because there isn't an existing option which is suitable.

My concern with that though is that it further forks this cookbook, and would be nice to unfork it one day...

lwgordonimos: On reflection, given we're forked anyway for now, it's not really a big deal. How about something like this?

*attributes/default.rb*
```
default['tomcat']['extra_java_options'] = '-Dhttps.protocols=TLSv1.2 -Djdk.tls.client.protocol=TLSv1.2'
```
https://github.com/aodn/chef/blob/master/cookbooks/tomcat/definitions/tomcat_instance.rb#L26
```
java_opts += " #{node['tomcat']['extra_java_options']}"
```
This way if a node really wanted to override it for some reason, it could just override *node['tomcat']['extra_java_options']*.

@anguss00 @jonescc ?

jonescc:
I was just looking at the title for the pull request.  If we want to add for all applications as a default then I'd add support for it in a cookbook but setting it on the node for only those nodes that need it is also good.

cookbooks/tomcat/attributes/default.rb
jonescc: Looks like we've completely hacked the tomcat cookbook so that looks like the best place.  Normally this kind of thing would go in a wrapper cookbook (generic_webapps for example)>

lwgordonimos: Yes, we were discussing that offline earlier, ideally we'd revert to upstream cookbook and wrap it, but for the moment we have this fork. By the time we get a chance to tackle that, the Java 7 workarounds should be a thing of the past...

PR #727, opened 2018-08-23
Use TLS 1.2 by default for Java applications

anguss00:


cookbooks/tomcat/attributes/default.rb
```diff
@@ -23,7 +23,7 @@
 default["tomcat"]["ports"]["ssl_port"] = 8443
 default["tomcat"]["ports"]["redirect_port"] = 8444
 default["tomcat"]["ports"]["ajp_port"] = 8009
-default["tomcat"]["java_options"] = "-Xmx128M -Djava.awt.headless=true"
+#default["tomcat"]["java_options"] = "-Xmx128M -Djava.awt.headless=true"
```
jonescc: Don't need to remove this?

anguss00:
Definitely @jonescc. That option was being overrided by us below but it was an unrelated change so I've removed it

PR #730, opened 2018-08-30
do not merge: removes chef-private from command post

gsatimos:


cookbooks/imos_command_post/templates/default/stackman_env.sh.erb
```diff
@@ -11,7 +11,6 @@ export AWS_DEFAULT_PROFILE=<%= @aws_default_profile %>
 # stackman parameter variables
 export STACK_NAME=aodnstack-${USER}
 export STACK_PROFILE="${HOME}/aodnstack-${USER}.yaml"
-export STACK_EBPREP_CONF_DIR=<%= @ebprep_conf_dir %>
```
lwgordonimos: We actually do need this set:

```bash
lwgordon@9-aws-syd:~$ stackman info
Usage: stackman info [OPTIONS]

Error: Missing option "-e" / "--ebprep-conf-dir".
```
Should probably be now set to in the recipe, and the environment var restored here:
```
File.join(repos_dir, 'cloud-deploy', 'sample-config', 'ebprep_conf')
```

akashisama: the deploy command uses a default ebprep_conf_dir value, info should really do the same thing

PR #732, opened 2018-09-03
update sumo logic default dev url

gsatimos:
@craigrose not a blocker for merging https://github.com/aodn/chef-private/pull/2980 but the default sumo logic collection url doesn't seem to exist in our account. The dev url in the account is currently 
```
REMOVED
```
so this is a PR to fix it... if you think that's correct? Perhaps someone hit the 'regenerate url' button


jonescc:
Actually, its not meant to exist in our account.  Its a url in a free account (which don't support metrics) so that everything works fine on the vagrant side.  I added this as the default for vagrant use so we don't impact metrics collection in our prod sumologic account - e.g. throttle metrics collection when playing around in vagrant (and not necessarily with metrics) and thuis losing production metrics.   Ideally we'd have a dev account - but that requires a paid subscription to get metrics at $324/month.

lwgordonimos:
Closing, as this should not be in chef.

PR #733, opened 2018-09-05
Revert "Update imos_postfix use use SES user for SMTP"

lwgordonimos:
Reverts aodn/chef#685


craigrose:
fixes https://github.com/aodn/issues/issues/279

PR #735, opened 2018-10-11
Add a 2.5.6 grails installation to jenkins for acoustics

akashisama:
due to grails version change in https://github.com/aodn/acoustic-data-viewer/pull/37


akashisama:
hmm actually the changes seem to work on 2.5.6 as well so maybe we should be upgrading to that...

akashisama:
Closing caz y'all replicated it in https://github.com/aodn/chef/pull/737

I did have a pink bear on the board 😛

PR #740, opened 2018-10-22
Update dependency orders in pipeline recipe.

lwgordonimos:

PR #742, opened 2018-11-01
Compliance checker/pip requirements update

lwgordonimos:
This is to enable the upgrade of the compliance-checker package.

1. lifts netCDF4 constraints due to conflicting with compliance-checker 4.1.1
1. ensures the 'libudunits2-dev' package is installed prior to installing Python requirements since it is a requirement for cfunits
1. includes/supersedes https://github.com/aodn/chef/pull/740


lwgordonimos:
@mhidas @bpasquer I've tested this in my PO box and it looks OK. Might be good if someone else tests in PO before we push to RC?

mhidas:
I'll give it a go.

mhidas:
Seems to work fine for me.

mhidas:
However I did have to do this first:  https://github.com/aodn/chef/pull/743

PR #757, opened 2019-01-10
Update squid template to only add configuration for webapps flagged a…

lwgordonimos:
…s cached


lwgordonimos:
@pmbohm noticed this bug/incorrect behaviour when thinking about the squid task.

PR #758, opened 2019-01-10
Poise-python 1.7.0 fix for pip 18.1

pmbohm:
Works for me from  afresh start. Still needs a few vagrant runs to fully provision


pmbohm:
fixes https://github.com/aodn/issues/issues/378

PR #762, opened 2019-02-12
install oracle 8 on jenkins and explicitly set system to openjdk 7

gsatimos:
this is part of talend java 8 upgrade task. 

This PR supports https://github.com/aodn/ci-config/pull/264 by installing oracle 8 jdk first (so that it's available as a jdk option)


cookbooks/imos_jenkins/recipes/managed_master.rb
```diff
@@ -133,4 +133,29 @@
   })
 end
 
+
+# install oracle 8 jdk
```
lwgordonimos: The "official" Java cookbook (which we have included anyway via the imos_java cookbook) should be able to handle all of this without resorting to `execute`:

https://supermarket.chef.io/cookbooks/java

Examples:
https://stackoverflow.com/questions/23815942/how-can-i-install-multiple-jdk-versions-using-opscode-java-cookbook
https://github.com/erichelgeson/jenkins-chef-dsl/blob/master/recipes/build-jdks.rb

gsatimos:
TOS 7 is the first version to support open jdk. closing this as probably won't need it

PR #763, opened 2019-02-13
Installs talend 7 in jenkins

gsatimos:
work in progress to install talend7 in jenkins alongside talend 5
part of https://github.com/aodn/backlog/issues/1054


gsatimos:
note that the TOS spatial extension that is currently being pointed to in s3 in the patched version. Once Craig J's submitted changes to upstream repo is built, we need to stop using the patched version

PR #764, opened 2019-02-19
Add sumologic CGI script to Nagios cookbook

lwgordonimos:
This forms a starting point for accepting passive Nagios checks from Sumologic.


lwgordonimos:
Nagios doco for passive checks: https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/passivechecks.html

Sumologic web hook doco: https://help.sumologic.com/Manage/Connections-and-Integrations/Webhook-Connections/Set-Up-Webhook-Connections

pmbohm:
todo:  Add a user for Sumo to Nagios

cookbooks/nagios/recipes/server.rb
```diff
@@ -422,8 +422,39 @@
...
 
+template "#{node['nagios']['config_dir']}/sumo.cfg" do
+  source "sumo.cfg.erb"
+  owner node['nagios']['user']
+  group node['nagios']['group']
+  mode 00440
+  variables({
+    :nagios_username => nagios_username,
+    :nagios_password => nagios_htpasswd,
+    :sumo_monitors => [
```
lwgordonimos: These need to be driven from a databag rather than inline here, but that shouldn't be too difficult.

craigrose: Is that not already the case as per (line 100) ?
```
  # In addition to users with htpasswd, we'll also add the fixed nagios user:
  nagios_username = Chef::EncryptedDataBagItem.load("passwords", "nagios")['username']
  nagios_htpasswd = Chef::EncryptedDataBagItem.load("passwords", "nagios")['htpasswd']
```

cookbooks/nagios/recipes/server.rb
```diff
@@ -422,8 +422,39 @@
   group node['apache']['user']
 end
 
+template "#{node['nagios']['config_dir']}/sumo.cfg" do
```
lwgordonimos: This should be replaced with the standard 'nagios_conf' resource, and sumo_monitors setup to be databag driven.

For example: 
https://github.com/aodn/chef/blob/master/cookbooks/nagios/recipes/server.rb#L337

Will need a new directory under data_bags and the relevant variables in the server.rb to load them and supply the values to the template.

PR #766, opened 2019-03-08
Use aodnfetcher for jenkins s3 hosted artifacts

anguss00:
Fixes https://github.com/aodn/issues/issues/391 by sourcing artifacts from S3 rather than Jenkins. Will require talend pipeline to archive in S3 only to complete fix.


cookbooks/imos_artifacts/libraries/fetcher_s3.rb
```diff
@@ -10,51 +10,52 @@
...
-          end
-        end
+        @artifact_filename = artifact_manifest['job']
+        # Assume wheel - TODO fix this assumption, aodnfetcher assumes .war file pattern
+        @artifact_uri = "jenkins://imos-binary/#{job_name}?pattern=^.*\\.whl$"
       end
+    end
 
-      @url = FetcherS3::get_job_url(bucket.url, job_name, buildNumber.to_s, @filename)
+    def download_and_return_path
```
lwgordonimos: Since the path is in the output JSON, so it would be simpler if this method just returned the parsed JSON document from *here*

cookbooks/imos_po/recipes/netcdf_checker.rb
```diff
@@ -11,6 +11,7 @@
 cchecker_artifact_path = Chef::Recipe::PipelineUtils.get_deployed_artifact_path(node['imos_po']['netcdf_checker']['cchecker_artifact'], node)
 ccplugin_artifact_path = Chef::Recipe::PipelineUtils.get_deployed_artifact_path(node['imos_po']['netcdf_checker']['ccplugin_artifact'], node)
 
+
```
lwgordonimos: blank line?

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -11,6 +11,8 @@
 require 'json'
 require 'pathname'
 
+include_recipe 'imos_po::netcdf_checker'
```
lwgordonimos: This shouldn't be in the pipeline recipe, since that recipe is for the system netcdf checker, and not related to fetching anyway.

private-sample/nodes/po.json
```diff
@@ -5,6 +5,7 @@
         "role[base]",
         "role[restore]",
         "recipe[ssl]",
+        "recipe[imos_core::aodnfetcher]",
```
lwgordonimos: I don't think this should be a separate standalone recipe, since it's just a dependency of the imos_artifacts cookbook, so that should just pull it in from there if it needs it, not make it the nodes responsibility to make sure it's explicitly installed.

cookbooks/imos_artifacts/libraries/fetcher_s3.rb
```diff
@@ -10,51 +10,52 @@
...
-      buildNumber = keys.map{|item| item.key().split('/')[2].to_i}.max
-
+      @artifact_filename = nil
       if artifact_manifest['filename']
-        @filename = artifact_manifest['filename']
-
+        @artifact_filename = artifact_manifest['filename']
+        artifact_prefix,artifact_suffix = @artifact_filename.split('.')
+        regex_pattern = "^.*#{artifact_prefix}.*\\.#{artifact_suffix}$"
+        @artifact_uri = "jenkins://imos-binary/#{job_name}?pattern=#{regex_pattern}"
```
lwgordonimos: Should pull out this hard coded bucket to an attribute.

cookbooks/imos_artifacts/libraries/fetcher_s3.rb
```diff
@@ -10,51 +10,52 @@
...
+        # Assume wheel - TODO fix this assumption, aodnfetcher assumes .war file pattern
+        @artifact_uri = "jenkins://imos-binary/#{job_name}?pattern=^.*\\.whl$"
       end
+    end
 
-      @url = FetcherS3::get_job_url(bucket.url, job_name, buildNumber.to_s, @filename)
+    def download_and_return_path
+      artifact_path = nil
+      if find_executable 'aodnfetcher'
+        system_cmd = "cd #{Chef::Config[:file_cache_path]} && aodnfetcher #{@artifact_uri} --cache_dir #{Chef::Config[:file_cache_path]}"
```
lwgordonimos: Way cleaner to get the fetcher to output a file itself, then you don't need to scrape standard output. The JSON.parse below can just take the file path then.

cookbooks/imos_artifacts/libraries/fetcher_s3.rb
anguss00: Yep, good point. I'll fix that up

anguss00:
@ready to go I think @lwgordonimos

anguss00:
Resource changes up @lwgordonimos to have a gander

cookbooks/imos_artifacts/libraries/fetcher_s3.rb
```diff
@@ -10,51 +10,46 @@
...
+    def download
+      artifact_path = nil
+      outfile = Tempfile.new('aodnfetcher')
+      output_path = outfile.path
+      system_cmd = "cd #{Chef::Config[:file_cache_path]} && aodnfetcher #{@artifact_uri} --outfile #{output_path} --cache_dir #{Chef::Config[:file_cache_path]}"
+      system(system_cmd)
+
+      begin
+        file = File.open(output_path, "rb")
+        contents = file.read
```
lwgordonimos: Still better to let the JSON library deal with the file object directly!

```
artifact_json = JSON.load(file)
```

cookbooks/imos_artifacts/providers/deploy.rb
```diff
@@ -30,15 +30,19 @@
...
-    Chef::Log.info("Deploying artifact '#{@cached_file_path}' -> '#{dest_file}' -> '#{install_dir}'")
-    ImosArtifacts::Deployer.extract_artifact(
-      @cached_file_path,
-      dest_file,
-      install_dir,
-      new_resource.owner,
-      new_resource.group,
-      new_resource.remove_top_level_directory)
-    new_resource.updated_by_last_action(true)
+    if ! @cached_file_path
```
lwgordonimos: This seems unrelated to the change, and looks like "workaround" code still.

cookbooks/imos_po/providers/deploy_wheel.rb
```diff
@@ -0,0 +1,35 @@
...
+# Copyright 2019, IMOS
+#
+# All rights reserved - Do Not Redistribute
+#
+# Provider to download pipeline python artifacts
+#
+
+attr_accessor :cached_file_path
+
+action :download do
```
lwgordonimos: 'download' is no longer accurate as the action name, probably should be 'deploy' or 'install'

cookbooks/imos_po/resources/deploy_wheel.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+
+actions :download
+default_action :download
+
+attribute :job_name, :kind_of => String, :required => true
+attribute :cached_file_path, :kind_of => String, :required => true
+attribute :venv_dir, :kind_of => String
+attribute :constraints_file, :kind_of => String
+attribute :python, :kind_of => String
```
lwgordonimos: Needs trailing newline

private-sample/nodes/po.json
lwgordonimos: ditto on previous omment

cookbooks/imos_core/attributes/aodnfetcher.rb
```diff
@@ -0,0 +1 @@
+default['imos_core']['aodnfetcher']['branch'] = 'master'
```
lwgordonimos: newline

cookbooks/imos_artifacts/recipes/default.rb
```diff
@@ -12,5 +12,7 @@
 # Needed to build ruby gems
 include_recipe "build-essential"
 
+include_recipe 'imos_core::aodnfetcher'
```
lwgordonimos: Just inline this, so we don't end up with more recipes just for dependencies

anguss00:
@lwgordonimos 
giphy (18)

PR #778, opened 2019-04-10
Created pipeline vagrant box for use with AWS resources

pmbohm:


pmbohm:
Test with http://thredds.aodn.org.au/thredds/catalog/IMOS/SRS/OC/BODBAW/2017_cruise-Lucinda_2017/absorption/catalog.html?dataset=IMOS/SRS/OC/BODBAW/2017_cruise-Lucinda_2017/absorption/IMOS_SRS-OC-BODBAW_X_20170207T035000Z_Lucinda_2017-absorption-CDOM_FV02_END-20171218T001100Z.nc

pmbohm:
In vagrant copy in the above file while watching the log

```
cp  /vagrant/shared_src/IMOS_SRS-OC-BODBAW_X_20170207T035000Z_Lucinda_2017-absorption-CDOM_FV02_END-20171218T001100Z.nc /var/incoming/SRS/OC/BODBAW/
tail -f /mnt/ebs/log/pipeline/process/tasks.SRS_OC_BODBAW.log
```

anguss00:
@lwgordonimos changes addressed

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -141,6 +141,11 @@
 pipeline_config = File.join(venv_etc_dir, 'pipeline.conf')
 pipeline_conf_databag = Chef::EncryptedDataBagItem.load('pipeline', node['imos_po']['pipeline']['pipeline_config_databag']).raw_data
 pipeline_conf = pipeline_conf_databag['config']
+
+if node['imos_po']['stack_integration'] != nil
+   pipeline_conf['global']['upload_uri'] = node['imos_po']['stack_integration']['bucket_name']
```
lwgordonimos: The URI consists of more than just the bucket name. Should be something like:

```ruby
pipeline_conf['global']['upload_uri'] = "s3://#{bucket_name}"
```

cookbooks/imos_po/templates/default/tasks.py.erb
```diff
@@ -39,6 +39,12 @@ end %>
 <%
 s3cmd_parts = Chef::Recipe::WatchJobs.get_s3cmd(node).split(" ")
 s3cmd_quoted = s3cmd_parts.map{ |s| "\"#{s}\""}.join(', ')
+
```
lwgordonimos: I'd leave this file out of the changes, as this is no longer used by anything and soon to be decomissioned

bin/pipeline-box.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
lwgordonimos: Can we roll this back into the other script? That supports different "modes" based on the name of the symlink, so would be better if we didn't duplicate/fork it here.

cookbooks/imos_po/recipes/pipeline.rb
pmbohm: thats wierd i have 
```
if node['imos_po']['pipeline']['upload_uri'] != nil
   pipeline_conf['global']['upload_uri'] = node['imos_po']['pipeline']['upload_uri']
end
```

lwgordonimos: This was changed to "single source" these values from the stack_integration node, so most likely happened as part of that. The idea being that we put the stack values in that one place, and everything reads/references from that rather than pasting the same dynamic value into multiple places.

cookbooks/imos_po/templates/default/tasks.py.erb
pmbohm: Voodoo I don't see any changes in my local branch since 2016. I may have to push force

gsatimos:
This PR appears to be part of the work for backlog item: https://github.com/aodn/backlog/issues/1090 which has a DoD on the board that I've picked up. Perhaps it is not actually ready for DoDing. 
Does anyone know the status?

cookbooks/imos_po/templates/default/tasks.py.erb
craigrose: @anguss00 I note these changes are yours.  Will you remove them?

cookbooks/imos_po/recipes/pipeline.rb
craigrose: The sample node['imos_po']['stack_integration']['bucket_name'] (pipeline,json) already contains the fully qualified s3 bucket name.  This seems better to me rather than hardcoding parts of the bucket name in the script.

PR #781, opened 2019-04-16
Make nagios do wfs checks on gml2 format

akashisama:
so we can pinpoint nagios requests in the geoserver controlflow plugin

(for https://github.com/aodn/backlog/issues/1096)

@lwgordonimos yeah this seems to take the same time as csv on problem layers so it should be safe


lwgordonimos:
This is why: https://osgeo-org.atlassian.net/browse/GEOS-4932

PR #782, opened 2019-04-16
Update Nagios Geoserver WFS check to validate feature count in XML ou…

lwgordonimos:
…tput


akashisama:
(for https://github.com/aodn/backlog/issues/1096)

PR #787, opened 2019-05-15
More detail in the po pipeline box readme.

craigrose:


pmbohm:
Cherry picking into #788 to be further modded

PR #801, opened 2019-06-06
Checking of AWS (EB and Stack) health on Nagios

pmbohm:


pmbohm:
Heres a screen shot of my last few hacks making some green colour!
Screen Shot 2019-06-06 at 2 54 05 pm

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -0,0 +1,93 @@
...
+        nagios_exit(2, message)
+
+    health = stack.get('StackStatus')
+    log.debug('Cloudformation  "{}" - status is {}'.format(stack_name, health))
+
+    nagios_exit(get_status_int(health), health)
+
+
+def get_status_int(status):
+    if 'COMPLETE' in status and status != 'DELETE_COMPLETE':
```
akashisama: ROLLBACK_COMPLETE and UPDATE_ROLLBACK_COMPLETE_CLEANUP_IN_PROGRESS should probably be treated as 2s. They suggest that although the stack is in a healthy state, an update that should have happened has failed.

cookbooks/nagios/templates/default/serverless.cfg.erb
```diff
@@ -32,7 +32,15 @@ define service {
 define service {
   service_description Elastic Beanstalk health status <%= vhost_name %>
   check_command check_eb!<%= "ci-eb-#{n['id']}-#{n['apptype']}" %>
-  host_name elastic-beanstalk-pseudo-host
+  host_name aws-pseudo-host
+  servicegroups serverless
+  use default-service
+}
+
+define service {
```
akashisama: We need to monitor all prod cloudformation stacks, not just the elastic beanstalk ones

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -0,0 +1,93 @@
...
+        return 0
```
akashisama: comments to show which statuses these integers match to would be helpful for readability

cookbooks/nagios/templates/default/serverless.cfg.erb
```diff
@@ -20,6 +28,22 @@ define service {
   servicegroups serverless
   use default-service
 }
+
+define service {
+  service_description Elastic Beanstalk health status <%= vhost_name %>
+  check_command check_eb!<%= "ci-eb-#{n['id']}-#{n['apptype']}" %>
+  host_name aws-pseudo-host
```
akashisama: would be cleaner to make this the same host as in the previous service for the websites

```suggestion
  host_name <%= n['id'] %>
```

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
lwgordonimos: You could also do it in a constant, to express the relationship between the string and the numerical value, e.g.

```python
STATUS = {
    "OK": 0,
    "WARNING": 1,
    "ERROR": 2,
    "UNKNOWN": 3
}

return STATUS["OK"]

```

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -0,0 +1,93 @@
...
+     'DELETE_COMPLETE','UPDATE_IN_PROGRESS','UPDATE_COMPLETE_CLEANUP_IN_PROGRESS','UPDATE_COMPLETE',
+     'UPDATE_ROLLBACK_IN_PROGRESS','UPDATE_ROLLBACK_FAILED','UPDATE_ROLLBACK_COMPLETE_CLEANUP_IN_PROGRESS',
+     'UPDATE_ROLLBACK_COMPLETE','REVIEW_IN_PROGRESS'
+    ]
+
+    print("\n\nDisplaying status label return values")
+    for status in statuses:
+        print("{} would return {}".format(status, get_status_int(status)))
+
+def nagios_exit(status, message):
```
lwgordonimos: Example of how this could look if the above status map was added:

```python
def nagios_exit(status_name, message):
    print("{}: {}".format(status_name, message))
    sys.exit(STATUS[status_name])

# calling code
nagios_exit("OK", "Nothing to see here, all ok.")
nagios_exit("ERROR", "Oh no, something wrong!")
```

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -0,0 +1,93 @@
...
+    for stack in cf_stacks.get('StackSummaries'):
+        print('Stack  "{}" is {}'.format(
+            stack.get('StackName'),
+            stack.get('StackStatus')
+        ))
+
+
+def check_cf_status(args):
+
+    stack_name = args[0]
```
akashisama: @lwgordonimos suggested in person that it would be good to use a library for dealing with the args so that if there's a problem it'll be dealt with cleanly at this point rather than later on

cookbooks/nagios/templates/default/serverless.cfg.erb
pmbohm: I'll need help on this one!

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
pmbohm: changed now to reflect your comments

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
pmbohm: yep ok

cookbooks/nagios/files/default/plugins/check_aws_eb_status.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+
+try:
```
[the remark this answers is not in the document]
pmbohm: I'll leave it here for testing

cookbooks/nagios/templates/default/serverless.cfg.erb
```diff
@@ -20,6 +28,22 @@ define service {
   servicegroups serverless
   use default-service
 }
+
+define service {
+  service_description Elastic Beanstalk health status <%= vhost_name %>
+  check_command check_eb!<%= "ci-eb-#{n['id']}-#{n['apptype']}" %>
```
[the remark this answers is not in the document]
pmbohm: @lwgordonimos Nice but i would need two hard coded variables that can easily be built following the convention. Are you anticipating exceptions to the `ci-eb` rule?



define service {
  service_description Elastic Beanstalk health status <%= vhost_name %>
  **check_command check_eb!<%= "ci-eb-#{n['id']}-#{n['apptype']}"** %>
  host_name <%= n['id'] %>
  servicegroups serverless
  use default-service
}

define service {
  service_description Cloudformation health status <%= vhost_name %>
  **check_command check_cf!<%= "ci-eb-#{n['id']}"** %>
  host_name <%= n['id'] %>
  servicegroups serverless
  use default-service
}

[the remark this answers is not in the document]
lwgordonimos: Probably not, but it makes sense to fully decouple this from any external conventions. The only important value is the environment name for check_eb OR the stack name for check_cf, so makes sense for Nagios to just take a list of names from databags rather than try to follow conventions at all.

pmbohm:
@akashisama @lwgordonimos 
Big review changes done

Requires https://github.com/aodn/chef-private/pull/3236 
*  Nagios checking of the newly defined nagios_cloudformation_stacks databag folder.
*  Elasticbeanstalk health checks are done from the nagios_serverlesshosts databag.

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
pmbohm: `args` is an array because my args are unnamed. Only the first one will be dealt with and only one is supplied by the Nagios config

lwgordonimos: http://nagios-plugins.org/doc/guidelines.html#AEN302

In a nutshell, the guidelines state that:

1. Use a good standard library
1. `Positional arguments are strongly discouraged.`

cookbooks/nagios/templates/default/serverless.cfg.erb
```diff
@@ -24,5 +27,18 @@ define service {
 <%
     end
   end
+
+  if n['hosttype'] == "eleasticbeanstalk"
```
akashisama:
```suggestion
  if n['hosttype'] == "elasticbeanstalk" 
```

cookbooks/nagios/templates/default/cloudformation_stacks.cfg.erb
```diff
@@ -0,0 +1,61 @@
...
+# Check AWS PROD Cloudformation stacks
+@cloudformation_prod_stacks.each do |n|
+
+  service_groups = n['service_groups'] ? n['service_groups'].join(',') : "serverless"
+  hostgroups_prod.push(n['hostgroups'])
+
+%>
+define service {
+  service_description Cloudformation Stack health status <%= n['id'] %>
+  check_command check_cf!<%= n['id']%>!AWS_PROFILE=production-developer;
```
akashisama: these won't really be developer profiles but specific to nagios so maybe "production-nagios" and "nonproduction-nagios" instead?

PR #804, opened 2019-06-12
Remove unused databag reference from PO box script

lwgordonimos:
This databag is not used, and creates an additional unnecessary step when deploying the PO box


lwgordonimos:
Yeah, I've made a separate PR. It was to access a private repo at one point in time, but that repo is now deleted.

anguss00:
delete

PR #808, opened 2019-06-12
S3 bucket exists nagios check

craigrose:
requires https://github.com/aodn/chef-private/pull/3244


cookbooks/nagios/files/default/plugins/check_s3.py
```diff
@@ -39,7 +43,26 @@ def parseArgs():
...
-nagios_exit("OK", "Bucket " + args.bucket_name + " exists")
+def check_host():
+    try:
+        s3 = boto3.resource('s3')
+        return s3
+    except:
+        nagios_exit("UNKNOWN", "Cannot connect to AWS S3")
+
+def main():
+    if not boto3:
```
akashisama: Pretty sure your assignments at line 13 and 14 mean that if either library fails to import then both will be None at this point.

craigrose: Yeah!  What on earth was I thinking?  Latest commit is more sensible.

PR #811, opened 2019-06-14
Aws batch nagios checks

craigrose:
requires https://github.com/aodn/chef-private/pull/3248


cookbooks/nagios/files/default/plugins/check_batch
```diff
@@ -0,0 +1,103 @@
...
+    return ivalue
+
+def parse_args():
+    argparser = argparse.ArgumentParser(description='Check AWS batch.  If one or more of -b, -f, -s is specified the checks '
+                                                    'will exit on first error or warning encountered.  Checks are made in'
+                                                    'the following order: failed jobs, blocked queue then queue size.')
+    argparser.add_argument("queue_name", help="Provide a job queue name.")
+    argparser.add_argument("-b", "--blocked_queue_time_limit", help="Hours a job is in RUNNABLE state before it triggers a warning.", type=check_positive)
+    argparser.add_argument("-f", "--failed_jobs_limits",
+                           help="[n m] First arg is the number of FAILED jobs that will trigger a warning.\nSecond arg is the number of FAILED jobs that will trigger an error.",
```
akashisama: I feel like it would be clearer to just have two seperate named arguments for the two trigger values

cookbooks/nagios/files/default/plugins/check_batch
```diff
@@ -0,0 +1,103 @@
...
+                           type=check_positive, nargs=2, default=[None, None])
+    argparser.add_argument("-s", "--queue_size_limit", help="Size of queue to trigger a warning", type=check_positive)
+    try:
+        args = argparser.parse_args()
+        return args
+    except SystemExit:
+        nagios_exit("UNKNOWN", "Error in check_batch arguments")
+        exit()
+
+args = parse_args()
```
akashisama: could you do this stuff in a main method a la @lwgordonimos's proposed standard at https://github.com/aodn/backlog/issues/1144#issuecomment-500691454

cookbooks/nagios/files/default/plugins/check_batch
craigrose: I did it this way because there must always be two values given for the checkFailed() check.  ie one on it's own doesn't make sense.

cookbooks/nagios/files/default/plugins/check_batch.py
```diff
@@ -0,0 +1,129 @@
...
+    runnable = batch_client.list_jobs(jobQueue=queue_name, jobStatus="RUNNABLE")["jobSummaryList"]
+    if len(runnable) > 0:
+        starting = batch_client.list_jobs(jobQueue=queue_name, jobStatus="STARTING")["jobSummaryList"]
+        running = batch_client.list_jobs(jobQueue=queue_name, jobStatus="RUNNING")["jobSummaryList"]
+        active_count = len(runnable) + len(starting) + len(running)
+        runnable.sort(key=createdAt)
+        runnable_age_of_oldest = (int(round(time() * 1000)) - runnable[0]["createdAt"]) // (1000 * 60 * 60)  # Hours
+        if runnable_age_of_oldest >= error_limit:
+            nagios_exit("ERROR", "Job(s) are blocked in queue: %s" % queue_name)
+        elif runnable_age_of_oldest >= warn_limit and active_count > 1:
```
akashisama: should use same logic as previous the if statment but with a different limit

```suggestion
        elif runnable_age_of_oldest >= warn_limit:
```

cookbooks/nagios/files/default/plugins/check_batch.py
```diff
@@ -0,0 +1,129 @@
...
+    try:
+        args = argparser.parse_args()
+        return args
+    except SystemExit:
+        nagios_exit("UNKNOWN", "Error in check_batch arguments")
+        exit()
+
+def check_host():
+    try:
+        boto3.setup_default_session(profile_name='production-developer')
```
gsatimos: @craigrose I saw in the Nagios (prod) log that all the batch checks return `Cannot connect to AWS Batch`
Perhaps it's this?

PR #813, opened 2019-06-17
Update recommended pg_basebackup command for more reliable restore

lwgordonimos:


lwgordonimos:
We haven't migrated anything to 10 yet, so 9.6.

craigrose:
ok we should change the --wal-method to --xlog-method then.  They are equivalent as far as I can see.

lwgordonimos:
Oops, quite right. I turned the short options into long options so it was clearer what they were doing, but added the v10 option name by mistake. I've amended the commit.

PR #815, opened 2019-06-18
add aws-wps api endpoint checks

craigrose:
requires https://github.com/aodn/chef-private/pull/3257


lwgordonimos:
It needs the `check_interval` to be configured at the service level: https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/4/en/objectdefinitions.html#service

This should be simple to add and retrieve from the databag, so that it gets templated in here: https://github.com/aodn/chef/pull/815/files#diff-abf6b9050f70b506d935ddbecfba99a8R19

PR #816, opened 2019-06-19
Install boto3 python library on nagios servers

akashisama:
so that our new aws service checks work


lwgordonimos:
As discussed, Poise Python is already responsible for managing pip, so we can get rid of the apt package install of pip. In fact we should get rid of any references to this anywhere else, but that's another PR!

PR #818, opened 2019-06-21
Move aws profiles setup from jenkins cookbook into imos_core

akashisama:
so we can use it for nagios as well (and in the future on other boxes like pipeline)

full disclosure: I haven't confirmed that this is working on 13-aws-syd just yet, because 13-aws-syd takes so long to cook, but putting it up so someone can get started on the review


cookbooks/imos_core/attributes/aws_profiles.rb
```diff
@@ -0,0 +1,4 @@
+default['imos_core']['aws_profiles']['aws_config_parent_folder'] = "~"
+default['imos_core']['aws_profiles']['mode'] = 00600
+default['imos_core']['aws_profiles']['user'] = "root"
+default['imos_core']['aws_profiles']['group'] = "root"
```
lwgordonimos: Trailing newline missing

cookbooks/imos_layer_seeder/templates/default/geowebcache.rb.erb
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/ruby
```
lwgordonimos: Unrelated file added to commit?

PR #826, opened 2019-06-24
Update status2xml.cgi with cleaner implementation to fix bug with che…

lwgordonimos:
…ck_command parameters with special characters

* non-templated Python, so it's independently testable
* inject Chef variables via environment variables rather than templating the script
* use library for building XML tree


anguss00:
Stunning

PR #831, opened 2019-06-26
Nagios checks for backlog#1140

pmbohm:
- lots of changes to bring into line with python standards
- handles cloudwatch stats for lambadas that are created in stacks with dynamic names and databags that dont know the full dynamic names


pmbohm:
Test with https://github.com/aodn/chef-private/pull/3271
For https://github.com/aodn/backlog/issues/1140

pmbohm:
closed https://github.com/aodn/chef/pull/833 as one place was needed to see the big picture

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
```diff
@@ -22,64 +22,133 @@
...
-        logging.error(msg)
-        nagios_exit("UNKNOWN", msg)
+        log.error(msg)
+        nagios_exit(STATUS["UNKNOWN"], msg)
 
-    args = parse_args()
-    cloudwatch = check_host()
-    check_metric(cloudwatch, args)
+    parser = argparse.ArgumentParser(description='Get a cloudwatch metric')
+    parser.add_argument("-q", "--query", required=True, help="Provide a cloudwatch query as a json string. Escape quotes.")
```
craigrose: query is a required argument, therefore should be positional not a '-q' option.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
```diff
@@ -22,64 +22,133 @@
...
+    parser.add_argument("-p", "--profile", type=str, help="AWS profile")
```
craigrose: I think all the argparse stuff in it's own function is cleaner and easier to read.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
```diff
@@ -22,64 +22,133 @@
...
 
-if __name__ == '__main__':
-    main()
+    args = parser.parse_args()
+    pp = pprint.PrettyPrinter(indent=4)
+    boto3_client = None
+
+    set_env()
+
+    check_metric()
```
craigrose: With this the script won't execute if it is run by importing. Although this is not our current intention the pattern from [here](https://github.com/aodn/backlog/issues/1144) is future proof.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
```diff
@@ -22,64 +22,133 @@
...
+
+def get_error_message(traceback):
+    error = traceback.format_exc().split("\n")
+    log.error(error)
+    return error[len(error)-2]
+
+
+def nagios_exit(status, message):
+    msg = STATUS.keys()[STATUS.values().index(status)]
+    print(msg + ": " + message)
```
craigrose: Just passing a status name string would be more direct and readable.  Otherwise use an enum.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
lwgordonimos: https://nagios-plugins.org/doc/guidelines.html#AEN302

`Positional arguments are strongly discouraged.`

craigrose: good to know!  nagios takes precedence but as a point of interest argparse says optional parameters should not be required (but they do provide for it)  :)

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
```diff
@@ -0,0 +1,162 @@
...
+        nagios_exit(STATUS["UNKNOWN"], msg)
+
+    parser = argparse.ArgumentParser(description='Get the last status of the AWS state machine')
+    parser.add_argument("-i", "--name", type=str, required=True, help="AWS state_machine identification <%= n['id'] %>:<%= n['service_name'] %> ")
+    parser.add_argument("-p", "--profile", type=str, help="AWS profile")
+    args = parser.parse_args()
+
+    set_env()
+
+    cloudformation_name, resource_type = args.name.split(":")
```
akashisama: Could these be retrieved as separate parameters by the argsparser rather than doing a string split? 

Also, note that these are being set as global variables, which I'd assume is bad practice. I suspect that's a reason why in @lwgordonimos's [standard](https://github.com/aodn/backlog/issues/1144#issuecomment-500691454) all this setup stuff is done in it's own method, "main()": that keeps the scope of all the variables clear.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
akashisama: Agree, but it should be added to the standard and other scripts if that's what we decide for this one.

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
pmbohm: @aodn/development-team Did you know Nagios itself is using positional arguments delimited by `!` when assembling the call to the script!

Ok although everything here works as its supposed to, I'll change everything again so it will survive being built and run in a full Jenkins pipeline and run as a lambda, 
oh and then write all the tests.

lwgordonimos: I think that's why they say don't write positional arguments on the scripts themselves. Get's way to hard to trace a parameter through if it's positional in two places, and the order *might* be different.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
```diff
@@ -22,64 +22,133 @@
...
-    cloudwatch = check_host()
-    check_metric(cloudwatch, args)
+    parser = argparse.ArgumentParser(description='Get a cloudwatch metric')
+    parser.add_argument("-q", "--query", required=True, help="Provide a cloudwatch query as a json string. Escape quotes.")
+    parser.add_argument("-p", "--profile", type=str, help="AWS profile")
 
-if __name__ == '__main__':
-    main()
+    args = parser.parse_args()
+    pp = pprint.PrettyPrinter(indent=4)
```
craigrose: PrettyPrinter
pp doesn't appear to be used.  As well as the import.

cookbooks/nagios/files/default/plugins/check_cloudwatch.py
```diff
@@ -22,64 +22,133 @@
...
 
-    args = parse_args()
-    cloudwatch = check_host()
-    check_metric(cloudwatch, args)
+    parser = argparse.ArgumentParser(description='Get a cloudwatch metric')
+    parser.add_argument("-q", "--query", required=True, help="Provide a cloudwatch query as a json string. Escape quotes.")
+    parser.add_argument("-p", "--profile", type=str, help="AWS profile")
 
-if __name__ == '__main__':
-    main()
```
craigrose: we should keep the pattern:
```
if __name__ == '__main__':
    main()
```

PR #833, opened 2019-06-27
Nagios checks of AWS StateMachines

pmbohm:
for https://github.com/aodn/backlog/issues/1140


pmbohm:
uses https://github.com/aodn/chef-private/pull/3273

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
```diff
@@ -0,0 +1,86 @@
...
+        msg = "Failed to load boto3 module! Is it installed?"
+        log.error(msg)
+        nagios_exit("UNKNOWN", msg)
+
+    parser = argparse.ArgumentParser(description='Get the last status of the AWS state machine')
+    parser.add_argument("-i", "--name", type=str, required=True, help="AWS state_machine identification <%= n['service_name'] %>:<%= n['id'] %> ")
+    parser.add_argument("-p", "--profile", type=str, required=True, help="cloudformation profile")
+
+    args = parser.parse_args()
+    boto3.setup_default_session(profile_name=args.profile)
```
akashisama: Better to do the boto3 set-up the same as in the cloudformation check

https://github.com/aodn/chef/blob/master/cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py#L25
``` python
def set_env(profile):

    global boto3_client
    try:
        boto_session = boto3.session.Session(profile_name=profile)
        boto3_client = boto_session.client('cloudformation')
    except:
        error = traceback.format_exc().split("\n")
        error_msg = error[len(error)-2]
        log.error(error_msg)
        nagios_exit(STATUS["UNKNOWN"], error_msg)

```
Two advantages I see: 
1. Good error handling, makes sure if something goes wrong there'll be a nice nagios error message
2. Explicitly creating a new session rather than changing the default session is good practice since it prevents code in other places from being inadvertently affected.

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
```diff
@@ -0,0 +1,86 @@
...
+    roleArnChunk, stateMachineArnChunk = args.name.split(":")
```
akashisama: Hmm, I don't at all like finding the resource via its IAM role, though I see why you had to do it: step functions evidently don't have any clean id or name to search for.

Instead I propose this:
1. Make the script accept the state machine ARN as a parameter
2. Add the state machine as a sub section of the "ci-lambda-tileseeder-prod-tileseeder" cloudformation stack databag
  - the subsection should contain the name of the state machine's cloudformation resource logical ID ("SeederStateMachine")
3. Make a script that finds out the ARN of a cloudformation resource give a cloudformation stack name and resource logical ID
4. Make the check_sm command discover the state machine ARN using the script from step 3, and then use that ARN to run the script from step 1

This will have the added advantage of using an existing host (the cloudformation stack) rather than creating a new one.

What do you think @pmbohm @lwgordonimos ?

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
```diff
@@ -0,0 +1,86 @@
...
+
+    return state_machines
+
+
+def get_state_machine_status_int(status):
+    """
+    :param status:
+    Valid Values: ACTIVE | DELETING
+    """
+    if 'ACTIVE' in status:
```
akashisama: This status doesn't tell us enough. Should also need to check the status of the last execution.

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
lwgordonimos: The only non-hacky way to find a dynamic ID is to query it from the *stack*. The stack should be the entry point for basically any checking of stack-managed resources, since that is the only truly constant thing in many cases.

e,g,
```python
import boto3

# these would come from parameters in a real solution
STACK_NAME = 'ci-lambda-aws-auto-off-dev-awsautooff'
RESOURCE_LOGICAL_ID = 'StateMachine'


cloudformation = boto3.client('cloudformation')

# list ALL resources in stack
all_resources = cloudformation.describe_stack_resources(StackName=STACK_NAME)['StackResources']

# find all state machines (you could keep this more dynamic and generic by just monitoring *all* found state
# machines in a stack, rather than finding one by name)
all_state_machines = [l for l in all_resources if l['ResourceType'] == 'AWS::StepFunctions::StateMachine']

# find the specific state machine based on the "logical" name (i.e. the name given in the template)
named_state_machine_resource = [l for l in all_state_machines if l['LogicalResourceId'] == RESOURCE_LOGICAL_ID][0]

# find the ARN from the resource dict
state_machine_arn = named_state_machine_resource['PhysicalResourceId']

print(state_machine_arn)


```

cookbooks/nagios/files/default/plugins/check_aws_statemachines.py
```diff
@@ -0,0 +1,86 @@
...
+
+    log.basicConfig(level=log.INFO,format='%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+    if not boto3:
+        msg = "Failed to load boto3 module! Is it installed?"
+        log.error(msg)
+        nagios_exit("UNKNOWN", msg)
+
+    parser = argparse.ArgumentParser(description='Get the last status of the AWS state machine')
+    parser.add_argument("-i", "--name", type=str, required=True, help="AWS state_machine identification <%= n['service_name'] %>:<%= n['id'] %> ")
+    parser.add_argument("-p", "--profile", type=str, required=True, help="cloudformation profile")
```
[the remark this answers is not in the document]
pmbohm: Why would we make this script not handle any profile? Even for testing?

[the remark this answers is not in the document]
lwgordonimos: If omitted, the library (and the various tools which use the library, e.g. awscli) loads the `default` profile.

What Alex is getting at is that `required=True` deviates from the "standard behaviour" in forcing the parameter to be provided.

pmbohm:
moving into one branch

PR #835, opened 2019-07-01
Only restart pipelines if databags change

craigrose:
This is intended to fix https://github.com/aodn/issues/issues/455.  It works, but not sure if it's the correct way?


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -10,6 +10,7 @@
 
 require 'json'
 require 'pathname'
+require 'fileutils'
```
lwgordonimos: I can't see why this is required? There is no new code that is doing anything unique (File operations and directory resources are already used).

craigrose: Removed it.  Was left over from earlier iteration where I used it to create the directory

PR #844, opened 2019-07-24
during talend7 installation, source codegen from new jenkins job

gsatimos:
re https://github.com/aodn/backlog/issues/1183


jonescc:
Looks good to me.  No reason to wait is there?   We just need to have deployed talend-codegen-edge and run it.

gsatimos:
correct, I already gave the talend_codegen_edge job a run yesterday and then tested 13-aws-syd in vagrant to make sure that it can source it okay. Alls good, so it's good to merge.

PR #846, opened 2019-07-25
New Dockerfile with packer and ChefDK installed

pmbohm:


private-sample/packer/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu:16.04
+
+ARG BUILDER_UID=9999
+
+ENV HOME /home/builder
+ENV BUCKET static-test-2.emii.org.au
```
lwgordonimos: This can be removed

private-sample/packer/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    wget awscli jq git perl unzip build-essential \
```
lwgordonimos: As far as I can see it should only need wget and unzip? Also, it's good to have these one per line for better diffs etc., e.g. https://github.com/aodn/python-aodndata/blob/master/Dockerfile#L6

private-sample/packer/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+# Install packer
+ENV PACKER_VERSION 1.4.2
+RUN mkdir -p /opt/packer
```
lwgordonimos: Even though it ends up looking weird, it's recommended that these related commands are run under a *single* run statement in the dockerfile, otherwise it screws up the layering stuff. For example, the RUN command should probably be to mkdir, wget, unzip and rm the zip all in one command.

See here for why:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

private-sample/packer/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+ENV HOME /home/builder
+ENV BUCKET static-test-2.emii.org.au
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    wget awscli jq git perl unzip build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install packer
+ENV PACKER_VERSION 1.4.2
```
lwgordonimos: This would be good as an ARG rather than an environment variable, since it's more of a parameter than a static variable

lwgordonimos:
As a result of discussions today, I've built upon this PR to make it represent a generic Chef/Packer image. This makes it suitable for offline use as well as use by Jenkins.

PR #857, opened 2019-08-01
change default talend version to 7

gsatimos:


gsatimos:
@jonescc could you please merge this so I can safely update notes (to remove the `version` param)

PR #860, opened 2019-08-02
Update list of pipeline dev nodes having editable pipeline packages i…

lwgordonimos:
…nstalled from local mount


lwgordonimos:
@craigrose I found this issue which wasn't picked up when getting rid of the old PO box.

Not having this would make pipeline dev infuriatingly overcomplicated to test changes...

PR #861, opened 2019-08-05
roll up cloudformation checks in nagios

gsatimos:
re https://github.com/aodn/issues/issues/483

Associated PR that needs to be merged with this PR: https://github.com/aodn/chef-private/pull/3317


cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -108,14 +129,14 @@ def main():
...
+    parser.add_argument("-i", "--instances", type=str, help="comma-separed list of cloudformation stack ids")
     parser.add_argument("-p", "--profile", type=str, help="cloudformation profile")
     args = parser.parse_args()
 
     set_env(args)
 
-    if args.instance:
-        check_cf_status(args.instance)
+    if args.instances:
+        check_cf_status(args.instances.split(','))
```
akashisama: it'd be good to strip white space from the stack names here, otherwise ```-p 'stack1, stack2'``` fails to check stack2 due to the space after the comma.

cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -108,14 +129,14 @@ def main():
         nagios_exit("UNKNOWN", msg)
 
     parser = argparse.ArgumentParser(description='A health checker of Cloudformation stacks')
-    parser.add_argument("-i", "--instance", type=str, help="cloudformation stack id")
+    parser.add_argument("-i", "--instances", type=str, help="comma-separed list of cloudformation stack ids")
```
akashisama: Should say "stack names" not "stack ids"

gsatimos:
@akashisama I've pushed an update.

I found that the method you suggested resulted in the second stack name given being treated as some unknown parameter (rather than part of the argument value that could be stripped of whitespace after the split).

I found a solution so that arg parse handles all the values after a flag as the intended value but with the downside that you need to actually use _spaces_ to separate the values.

I guess then it makes it unambiguous to use but with the downside that 
- using comma-separated lists are recommended in nagios guidelines http://nagios-plugins.org/doc/guidelines.html
- I can't see any other example of any of our check commands using a space separation.

What do you reckon?

akashisama:
I think it's better to keep the comma separated style to keep it in line with the other checks.

Turns out my ```-p 'stack1, stack2'``` suggestion was wrong and should have been ```-p "stack1, stack2"``` which is what I was doing on my machine. I didn't realise the different types of quotation marks did different things! With double quotes around the csv python interprets it as two variables "stack1" and " stack2" with a space at the start.  

```check_cf_status(name.lstrip() for name in args.instances.split(','))``` works for the white space stripping once double quotes are used.

**Comment by gsatimos** [comment]
yep, that method seems best. update pushed

File: cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py [hunk 3]
```diff
@@ -108,14 +129,14 @@ def main():
         nagios_exit("UNKNOWN", msg)
 
     parser = argparse.ArgumentParser(description='A health checker of Cloudformation stacks')
-    parser.add_argument("-i", "--instance", type=str, help="cloudformation stack id")
+    parser.add_argument("-i", "--instances", type=str, help="space-separated list of cloudformation stack names")
```
akashisama: need to change back the help message here

gsatimos:
sorted

lwgordonimos:
The "instances" terminology is pretty confusing here... EC2 has "instances", Cloudformation has "stacks".

gsatimos:
agreed. I've pushed an update to correct this.

PR #866, opened 2019-08-20
experiment in restarting watches

craigrose:
re: https://github.com/aodn/issues/issues/490
Need some help with this.  Trying to make sense of how this can work with subscribes.  My goal is to trigger restarts through subscribing to imos_deploy_wheel[cached_file_path] (imos_deploy_wheel[cached_file_path_dummy] for testing).

But I don't get any log messages indicating restarts - nor changes in the pids for the workers.
My first suspicion is that the array watches_deployed does not persist / or not visible to chef outside the ruby block? 

Appreciate any help with this when you can. Thanks.


lwgordonimos:
Have a look at this... specifically, the section regarding `updated_by_last_action`. Doesn't look to me like it's setting the updated attribute anywhere, so probably not registering as changed?

https://docs-archive.chef.io/release/12-0/custom_resources.html

PR #868, opened 2019-08-22
Fix renamed ncwriter repo

pmbohm:
Fixes aodn/issues/issues/509

The AODN repo `aodn-netcdf-tools` was renamed to `python-aodntools`
Fixes Vagrant pipeline box runs. 
#### Note
1. Vagrant users will most likely need to clear out their chef `src` directory, also mounted in the Vagrant box at `/vagrant/src`. Chef dosen't seem to Git pull on these locally checked out repos?
1. `/etc/init.d/supervisor start`fails as the directory  `/vagrant/src/data-services/log/celery` has not been created


bin/vagrant-wrapper.sh
```diff
@@ -186,7 +186,7 @@ po_box7_main() {
     declare -r DATA_SERVICES_GIT_REPO=git@github.com:aodn/data-services
     declare -r AODNCORE_GIT_REPO=git@github.com:aodn/python-aodncore
     declare -r AODNDATA_GIT_REPO=git@github.com:aodn/python-aodndata
-    declare -r NCWRITER_GIT_REPO=git@github.com:aodn/aodn-netcdf-tools
+    declare -r NCWRITER_GIT_REPO=git@github.com:aodn/python-aodntools
```
lwgordonimos: Looks good, but would be good to also rename this, for consistency (also the other reference to NCWRITER_GIT_REPO below)

PR #872, opened 2019-08-29
Get Jenkins V2 harvesters using s3prefix fetcher

anguss00:
in the talend databag:
``file//vagrant/src/AUV_harvester_Latest.zip`` will source the file from ``/vagrant/src/AUV_harvester_Latest.zip``

``jenkins/systest/AUV_harvester_Latest.zip`` will source the file from the systest promoted harvester ``harvester_<id>``


cookbooks/imos_artifacts/libraries/fetcher_s3_prefix.rb
```diff
@@ -0,0 +1,65 @@
...
+
+require 'json'
+require 'tempfile'
+
+module ImosArtifacts
+  class FetcherS3Prefix
+
+    def initialize(artifact_manifest, node)
+      job_name = artifact_manifest['job']
+      if node['chef_environment'].eql?('production') || node['chef_environment'].eql?('ci')
```
akashisama: This is right for harvesters but will need to be changed when we start using the prefix fetcher to fetch webapp artifacts as well. 

I feel like it would be better to set the environment in the artifact manifest. This has the advantage that it will be easy to set for non-harvester artifacts in their node definitions, and can be set for harvester artifacts at https://github.com/aodn/chef/pull/872/files#diff-a1130f5961683ecfe91a4cce0d08d8a8R44 where the artifact manifest is already being manipulated.

cookbooks/imos_artifacts/libraries/deployer.rb
```diff
@@ -53,6 +53,10 @@ def self.is_uri(uri)
       return uri.start_with?("/") || uri.start_with?("http://") || uri.start_with?("https://")
     end
 
+    def self.is_harvester_job(job)
```
lwgordonimos: This looks like it should be removed now?

cookbooks/imos_artifacts/libraries/fetcher_s3_prefix.rb
lwgordonimos: The chef_environment is nothing to do with the *requested artifact*, so doesn't make any sense to couple them like this. The calling code will ask for an artifact by name, so the only appropriate behaviour here is to give it the artifact that it asked for, not perform any logic to control which artifact it gets. The job name AND the environment need to come through as a parameter somehow.

cookbooks/imos_artifacts/libraries/fetcher_s3_prefix.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+      @artifact_filename = nil
+      if artifact_manifest['filename']
+        @artifact_filename = artifact_manifest['filename']
+        artifact_prefix,artifact_suffix = @artifact_filename.split('.')
+        regex_pattern = "^.*#{artifact_prefix}.*\.#{artifact_suffix}$"
+        @artifact_uri = "s3prefix://#{node[:imos_artifacts][:s3_prefix_bucket]}/promoted/#{job_name}/#{environment}?pattern=#{regex_pattern}"
+      else
+        @artifact_filename = artifact_manifest['job']
+        regex_pattern = '^\\(\\(?\\!md5\\).\\)*$'
```
lwgordonimos: md5s will never be relevant on the new bucket, so no need to have special cases for them

anguss00:
@lwgordonimos @akashisama comments addressed

cookbooks/talend/recipes/default.rb
```diff
@@ -42,7 +42,22 @@ def get_job_common_name(job_name, data_bag_item)
...
   environment = node['talend']['environment']
 
-  bag_item = data_bag_item("talend", job_name)
+  if node['chef_environment'].eql?('production') || node['chef_environment'].eql?('ci')
+    deployment_stage = 'production'
+  else
+    deployment_stage = 'systest'
+  end
+
+  if job_name.include?("s3prefix")
```
lwgordonimos: The name "s3prefix" is an internal detail, so doesn't make sense for that naming to propagate around. This is what appears in the node files, so from the perspective of somehow editing those, "s3prefix" is pretty meaningless. All they need to know that they need to put in a local path, and some intuitive string which indicates that it should come from jenkins.

PR #877, opened 2019-09-05
Deploy wheel changes

lwgordonimos:


lwgordonimos:
@craigrose I've been working on this today, and think it's almost there.

I've put a test commit on the end with the super cut down node and cookbook I've been using for testing, which we will need to get rid of before merging. Needs more testing in the context of the whole pipeline recipe, and probably some more output to indicate what the runtime values are and why it is or isn't restarting the services.

craigrose:
Looks like where I was heading and answers the two problems I was having.  A good pip command to get the version and I was missing exactly what updated_by_last_action was doing.  I'll do the testing and sort out the logging/output.

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -245,27 +245,37 @@
...
     job_name node['imos_po']['pipeline']['aodndata_artifact']
     constraints_file constraints_file
     venv_dir environment_dir
   end
+
+  restart_all_services = [
+      aodncore.updated_by_last_action?,
+      aodntools.updated_by_last_action?,
+      aodndata.updated_by_last_action?
+  ].any?
```
craigrose: restart_all_services will only be evaluated during chef compile phase and will always be false.  We need to evaluate it as part of a resource during converge.

lwgordonimos: Yeah sounds good, I hadn't really tested this, as I was mostly focussing on the under the hood change detection stuff in the dummy cookbook. There may be other bits that need to be moved into a resource too.

cookbooks/imos_po/providers/deploy_wheel.rb
```diff
@@ -12,24 +12,46 @@
...
+    # Download wheel file for installation candidate
     artifact_manifest = ImosArtifacts::Deployer.get_artifact_manifest(new_resource.job_name)
     cached_file_path = ImosArtifacts::Fetcher.new.fetch_artifact(artifact_manifest, node)
 
-    if new_resource.constraints_file
-      command = "-m pip install -c #{new_resource.constraints_file} #{cached_file_path}"
-    else
-      command = "-m pip install #{cached_file_path}"
-    end
+    # Get new package information
```
lwgordonimos: I think all this logic would be better in a supporting Ruby class, similar to `ImosArtifacts::Deployer.get_artifact_manifest` above.

PR #878, opened 2019-09-17
 Update to  Python3

pmbohm:
...  with: python-modernize -w -p

for aodn/backlog#1275

PR #883, opened 2019-09-25
Revert " Update to  Python3"

akashisama:
Reverts aodn/chef#878


akashisama:
Reverted because it wasn't ready to go into rc (which runs off master branch).

PR #886, opened 2019-10-04
rewrite config before restarting each watcher

craigrose:
I've made this a draft for discussion then we'll see if it cuts the mustard!
- First up, I noticed that we were actually restarting the watchers (and other services) twice each run.  Once during compile phase and then during converge.  Running during compile may have the potential to restart before the config is ready.  https://github.com/aodn/chef/blob/b9cb9a4906c4119df8d2c825bd51e6c0b88ab681/cookbooks/imos_po/recipes/pipeline.rb#L464
- Whilst trying to implement a notify or subscription on the config file changes I found that subscriptions were ignored when inside the ruby_block `block do` section.  I did find docs on how to notify other resources which are outside the ruby_block but this is not what we want here.  There were two places where we are trying to subscribe from with the `block do` and I could not see evidence that these were actually doing anything - so I removed them. 
- Finally I slightly reframed the problem as "make sure correct config is in place before restarting each watcher". My solution to this is to create the watch config file empty before the loop and update it each time we detect it will change just before restarting the service.


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -437,18 +442,12 @@
       f.run_action :enable
       if watch_config_changed or restart_all_services
         Chef::Log.info "Restarting #{service_name} service due to code change (#{restart_all_services}) or configuration change (#{watch_config_changed})"
+        wc.content Chef::JSONCompat.to_json_pretty(watch_config_hash)
+        wc.run_action :create  # Updates the file with new config for this pipeline
         f.run_action :restart
       end
-      f.subscribes :create, "directory[#{watch_log_dir}]"
```
craigrose: Removed because it does seem to do anything

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -458,9 +457,9 @@
     pws.user watch_service_user
     pws.run_action :enable
     pws.run_action :restart
-    pws.subscribes :create, "directory[#{watchservice_log_dir}]"
```
craigrose: Removed because it does seem to do anything

PR #889, opened 2019-10-16
Update pipeline box documentation

mhidas:
* Fix URL used to check content of dev stack S3 bucket (See https://github.com/aodn/issues/issues/553#issuecomment-541981509)
* Add a note on copying stack_integration details into the pipeline box config (I wasted some time failing to do this on first try)


pmbohm:
@mhidas Is it worth adding the protocol prefixes to --show-secrets ?

mhidas:
> @mhidas Is it worth adding the protocol prefixes to --show-secrets ?

Yes, that could be helpful I guess.

PR #890, opened 2019-10-17
link anmn_am nrt product to correct metadata uuid for pipeline box use

evacougnon:
Needs to be merged at the same time than the [PR in harvesters](https://github.com/aodn/harvesters/pull/788) 
Also, better to wait for [this issue](https://github.com/aodn/issues/issues/555) to be solved, so we are able to fully test it in the pipeline box


mhidas:
Ready for testing in pipeline box!

PR #892, opened 2019-10-21
Pipeline readme additions

pmbohm:


doc/README.pipeline-box.md
```diff
@@ -202,17 +202,13 @@ Then you need to deploy the harvester:
...
 
-By default, po box can harvest from https://catalogue-123.aodn.org.au/geonetwork
+## Importing Geonetwork Records
 
-Import vocabularies and load and initiate the harvester by running:
-```
-$ vagrant ssh po -- /var/lib/tomcat7/default/geonetwork-config-manager.rb -g http://localhost:8080/geonetwork
-```
+By default the Geonetwork has no imported metadata records, although all of the metadata harvesters are in place.
+Initiate the harvester of your choice by logging in to https://`[your stack geonetwork]`.dev.aodn.org.au/geonetwork/srv/eng/harvesting
```
akashisama:
```suggestion
Initiate the harvester of your choice by logging in to http://geonetwork-aodnstack-${USER}.dev.aodn.org.au/geonetwork/srv/eng/harvesting
```

Makes it consistent with how we refer to stack endpoints earlier in the document, eg line 45:

```
1. Confirm that expected objects have been included in the S3 bucket at `http://imos-data-aodnstack-${USER}.dev.aodn.org.au
````

Also, geonetwork endpoint is http not https.

File: doc/README.pipeline-box.md [hunk 2]
```diff
@@ -202,17 +202,13 @@ Then you need to deploy the harvester:
...
 
-Import vocabularies and load and initiate the harvester by running:
-```
-$ vagrant ssh po -- /var/lib/tomcat7/default/geonetwork-config-manager.rb -g http://localhost:8080/geonetwork
-```
+By default the Geonetwork has no imported metadata records, although all of the metadata harvesters are in place.
+Initiate the harvester of your choice by logging in to http://geonetwork-aodnstack-${USER}.dev.aodn.org.au/geonetwork/srv/eng/harvesting
 
 Notes:
-* Vocabularies are required for facetted searching to be available in the portal
```
mhidas: Why are you removing this? Vocabs _are_ required for facetted searching, and it seems they are not available in the stack, except for Organisations. Is that as intended?

doc/README.pipeline-box.md
```diff
@@ -202,17 +202,13 @@ Then you need to deploy the harvester:
...
 * The imported records will link to production online resources (i.e. layers in geoserver-123), not the ones from the box itself. If you want to test layers from data in the box, you need to update the links in the relevant record.
```
mhidas: Strangely, I just ran the IMOS harvest, and _some_ online resources in the records are updated to point to the stack geoserver/geowebcache. The only one that still points to geoserver-123 is the one with protocol "AODN:WFS-EXTERNAL-1.0.0-http-get-capabilities".

doc/README.pipeline-box.md
pmbohm: Vocabs are loaded via cloud_deploy. 

The work has just been done to make this compatible for the Geonetwork3 API following the existing code for Geonetwork2

doc/README.pipeline-box.md
pmbohm: To be honest I haven't run a harvest myself nor am i knowledgable whether a metadata transform is correctly applied? I erred on the side of caution to leave this line in.
@lwgordonimos ?

doc/README.pipeline-box.md
pmbohm: To clarify ` stackman init ` on `commandpost.aodn.org.au` creates a YAML file that defines the vocabs below to install.
Is this not working?

```
  vocabularies:
    - aodn_instrument
    - aodn_organisation
    - aodn_organisation_category
    - aodn_parameter_category
    - aodn_platform
    - aodn_platform_category
    - aodn_sampling_parameter
    - aodn_units_of_measure
    - australian_discovery_parameter
    - land_masses
    - region
    - water_bodies
```

doc/README.pipeline-box.md
lwgordonimos: It's a slightly modified version of the systest transform. There shouldn't be much magic to it, it should just be running this:

https://github.com/aodn/schema-plugins/blob/master/iso19139.mcp-2.0/process/ebstack-linkage-updater.xsl

pmbohm: I'll rip out this line too then

doc/README.pipeline-box.md
mhidas: As far as I can tell from the GN interface, they're all there:
Screenshot from 2019-10-22 17-30-48
... but there's nothing under the facets in the portal. I'll lodge that as a separate issue.

doc/README.pipeline-box.md
mhidas: As I said, there seems to be one online resource link in each record that has not been transformed. Will lodge separate issue.

mhidas: Actually I'm guessing this is by design. The WFS link in each record is duplicated. There is one with the usual "OGC:WFS-1.0.0-http-get-capabilities" protocol, pointing at the stack geoserver, and a copy of it, with protocol "AODN:**WFS-EXTERNAL**-1.0.0-http-get-capabilities", pointing at geoserver-123.
Who would be the best person to confirm this (and save me logging yet another issue)?

doc/README.pipeline-box.md
```diff
@@ -202,17 +202,14 @@ Then you need to deploy the harvester:
...
-$ vagrant ssh po -- /var/lib/tomcat7/default/geonetwork-config-manager.rb -g http://localhost:8080/geonetwork
-```
+By default the Geonetwork has no imported metadata records, although all of the metadata harvesters are in place.
+Initiate the harvester of your choice by logging in to http://geonetwork-aodnstack-${USER}.dev.aodn.org.au/geonetwork/srv/eng/harvesting
 
-Notes:
-* Vocabularies are required for facetted searching to be available in the portal
-* The imported records will link to production online resources (i.e. layers in geoserver-123), not the ones from the box itself. If you want to test layers from data in the box, you need to update the links in the relevant record.
+#### Further Harvesting information
+https://geonetwork-opensource.org/manuals/2.10.4/eng/users/managing_metadata/harvesting/index.html#the-main-page
```
mhidas: :+1:

doc/README.pipeline-box.md
```diff
@@ -202,17 +202,14 @@ Then you need to deploy the harvester:
...
 
+#### Notes:
 Before you can edit the harvested records, you need to go into Geonetwork's **System configuration** (under **Catalogue settings** on the **Administration** page) and tick the "Allow editing on harvested records" option near the bottom of the page.
```
mhidas: Just one last question: does the "Allow editing on harvested records" option need to be enabled before a harvester can update the spatial extent in the record? If so, it would be good to explicitly point this out too. Otherwise this may cause some unnecessary head-scratching when harvesters fail because they can't edit the record.

pmbohm: Well i guess this is the case from what it says here. 
What used to happen? The text is unchanged.
It is also the case that there might be many more nips and tucks that might be found as we go along

mhidas: Ok, we'll leave it.

doc/README.pipeline-box.md
lwgordonimos: It's likely by design, looks to be as a result of trying to make Geoserver internal.

https://github.com/aodn/aodn-portal/pull/2565

I wouldn't change it at this stage unless it's preventing something in the stack from working, otherwise it might be a simple fix in the transform.

PR #898, opened 2019-10-29
Use new jenkins artifacts in po7 box and systest env

anguss00:


cookbooks/imos_po/attributes/pipeline.rb
```diff
@@ -14,10 +14,10 @@
 default['imos_po']['pipeline']['local_aodntools_repo'] = '/vagrant/src/python-aodntools'
 
 # Jenkins job names for Python package artifacts
-default['imos_po']['pipeline']['aodncore_artifact'] = 'aodncore_prod'
-default['imos_po']['pipeline']['aodndata_artifact'] = 'aodndata_prod'
-default['imos_po']['pipeline']['aodntools_artifact'] = 'aodntools_prod'
-default['imos_po']['pipeline']['cc_plugin_imos_artifact'] = 'cc_plugin_imos_prod'
+default['imos_po']['pipeline']['aodncore_artifact'] = 'jenkins/production/python-aodncore'
```
lwgordonimos: All these updates to the defaults should be in a separate PR, because it will affect cooking prod nodes during the test period.

anguss00:
@akashisama updated to separate PO box changes

PR #907, opened 2019-11-07
Make Python 3 the default for the Pipeline 2 virtualenv

lwgordonimos:
Since production is now running 3, we can make Python 3 the default and remove the overrides.


akashisama:
https://github.com/aodn/backlog/issues/1343

PR #911, opened 2019-11-18
Install gdal from pypi

anguss00:
https://github.com/aodn/data-services/pull/963


anguss00:
PR 911

porsche

cookbooks/imos_po/recipes/packages.rb
```diff
@@ -26,6 +26,12 @@
   action :upgrade
 end
 
+python_package 'pip' do
```
lwgordonimos: This is just repeating the above resource and isn't related to GDAL, so should be removed

cookbooks/imos_po/recipes/packages.rb
```diff
@@ -26,6 +26,12 @@
...
+  python python '/usr/bin/python3'
+  version node['imos_po']['data_services']['pip_version']
+  action :upgrade
+end
+
 python_package 'pyopenssl' do
```
lwgordonimos: There needs to be another resource to ensure that the old package is removed e.g.

```ruby
apt_package 'imos-gdal' do
  action :remove
end
```

PR #919, opened 2019-12-01
Split Nagios checks of jenkins

pmbohm:
For aodn/backlog#1355
requires aodn/chef-private#


cookbooks/nagios/files/default/plugins/check_jenkins
```diff
@@ -57,7 +57,12 @@ def get_jenkins_job_status(opts)
     return 3
   end
 
-  job_color = job_status['color']
+  if job_status['_class'] == "org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject"
+    job = job_status['jobs'].select {|item| "master" == item['name']}
```
akashisama: This won't work for thredds and geonetwork since their master branch isn't called 'master'. Would it be possible to list the lower level job directly in the json maybe? Like:
``` JSON
 "monitored_jobs": {
    "geonetwork-build_build:2.10.x-imos": []
}
```

or something like that?

pmbohm: I guess this would be possible to extend the script. Is the colon a safe delimiter?

akashisama: Looks like the following url works for getting status of lower level job: http://build.aodn.org.au/job/core-geonetwork_build/job/2.10.x-imos/api/json

So could probably do this:
``` JSON
 "monitored_jobs": {
    "core-geonetwork_build/job/2.10.x-imos": []
}
```

akashisama: > I guess this would be possible to extend the script. Is the colon a safe delimiter?
The script is making a url out of it, so if we provide a job name that can slot into line 50 as part of the url that should work I think

lwgordonimos: @akashisama @pmbohm see related PR (and my other comment). Nagios shouldn't need logic to deal with this.

lwgordonimos: Ok, had some good chats with Phil, so disregard that re: the weather column. The weather column isn't really useful for this, since it takes old builds into account, and there's no point alerting on build failures that aren't happening anymore. I think what you've got here is the best method in terms of Nagios.

PR #920, opened 2019-12-02
Rename artifact_filename to artifact_uri

pmbohm:
Fixes https://github.com/aodn/issues/issues/597


anguss00:
We will also need to update the paths for the URIs to fit into the new pattern for harvesters. It could be best to just copy the databags from private to replicate the config there?

anguss00:
See https://github.com/aodn/chef-private/commit/ef6d846765d57b74097f5d8f39e5951b566587d9

PR #921, opened 2019-12-02
Update readme with local artifact changes

anguss00:


pmbohm:
Theres no reference to the ability to load from S3 ?

anguss00:
Added in @pmbohm

PR #922, opened 2019-12-03
Create explicit pipeline_requirements.txt file

pmbohm:
Fixes https://github.com/aodn/issues/issues/598
Fixes https://github.com/aodn/issues/issues/599

Locks the versions of certain packages to stop the various issues.


cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -202,48 +202,52 @@
...
+cookbook_file pipeline_requirements do
+  source 'pipeline_requirements.txt'
   action :create
 end
 
+pip_requirements pipeline_requirements do
+  virtualenv environment_dir
+end
+
+pip_requirements File.join(node['imos_po']['pipeline']['local_aodncore_repo'], 'requirements.txt') do
```
lwgordonimos: This looks like it was copy/pasted from below and left here?

cookbooks/imos_po/recipes/pipeline.rb
```diff
@@ -202,48 +202,52 @@
   virtualenv environment_dir
 end
 
-constraints_file = File.join(Chef::Config[:file_cache_path], 'pip_constraints.txt')
-cookbook_file constraints_file do
-  source 'pip_constraints.txt'
+pipeline_requirements = File.join(Chef::Config[:file_cache_path], 'pipeline_requirements.txt')
```
lwgordonimos: Again, despite my previous comment to the contrary, I think we actually *should* leave the constraints.txt in now I look at it. I think that gives us maximum control, since it also would prevent subsequent packages from installing requirements which would override the virtualenv version. If we want a hard cap for some reason (for example in the current case of python-dateutil where we know it just won't work) the constraints may still be the best way to ensure that cap is enforced.

pmbohm:
@lwgordonimos review comments addressed

cookbooks/imos_po/files/default/pipeline_requirements.txt
```diff
@@ -0,0 +1,8 @@
+compliance-checker==4.1.1
```
lwgordonimos: @mhidas  this is the current thinking around managing the dependency issues in the pipeline virtualenv. Rather than trying to duplicate pinned versions in each of our Python packages, we define the common dependency versions here. Then we can move the per-package requirements back to simply *compatible* requirements (e.g. ranges of versions or greater than version X) rather than "specific version requirements we actually have in the virtualenv".

Interested in your thoughts on this idea.

mhidas: That makes sense to me. I assume this will then be the same environment deployed to pipeline box, RC, and prod?

The obvious (slight) drawback is e.g. if we add some code in aodndata that now requires pandas 0.25, it's not enough to update the requirements in the aodndata package, we also have to update it here. However that shouldn't happen too often.

PR #923, opened 2019-12-03
Vagrant essentials to stop wasting time

pmbohm:


Vagrantfile
```diff
@@ -65,4 +65,10 @@ Vagrant.configure("2") do |config|
     collect(&basename).
     each(&define_node)
 
+  config.vm.provider "virtualbox" do |vb|
+    vb.memory = "4096"
```
lwgordonimos: Memory is usually set by the VAGRANT_MEMORY environment variable, so this shouldn't override that. Maybe an if statement to only run this if the variable isn't set.

PR #929, opened 2019-12-09
Constrain pipeline jsonschema version to < 3.0

mhidas:
Versions of jsonschema above 3.0 break aodntools. This limit is in the [aodntools setup.py](https://github.com/aodn/python-aodntools/blob/master/setup.py#L4), but I think we also need it here, to prevent chef from upgrading this package after installing aodntools.


mhidas:
Ok, this is not needed after all.

PR #933, opened 2019-12-18
Fixes aodn/issues#564 using pagination

pmbohm:


cookbooks/nagios/files/default/plugins/check_aws_cloudformation.py
```diff
@@ -53,17 +53,22 @@ def check_cf_statuses():
...
+        ds_paginator = boto3_client.get_paginator('describe_stacks')
+        response_iterator = ds_paginator.paginate()
+        for items in response_iterator:
+            for stack in items['Stacks']:
+                if stack['StackName'] in stack_names:
+                    stack_info.append(
+                        {
+                            'name': stack.get('StackName'),
+                            'status': get_stack_status(stack.get('StackStatus'))
+                        })
```
akashisama: I think this would be more pythony: 

``` Python
    stack_info = [
        {
            'name': stack.get('StackName'),
            'status': get_stack_status(stack.get('StackStatus'))
        } for stack in items['Stacks'] if stack['StackName'] in stack_names]
```

(since it does the "for" and "if" in one line and python people love that)

pmbohm: Yeah it took me a little unpicking to understand so it must be the right way.

pmbohm:
@akashisama Pythony-ated

lwgordonimos:
https://leadsift.com/loop-map-list-comprehension/

TL;DR. if you're making a list, use a list comprehension. If you're not making a list, use a loop.

PR #951, opened 2020-02-12
vsftpd configuration for FTPS

anguss00:


cookbooks/imos_vsftpd/templates/default/vsftpd.conf.erb
```diff
@@ -18,20 +18,26 @@ pam_service_name=pam_vsftpd
...
 guest_enable=YES
 guest_username=ftp
-# This is important
+# SSL settings
+<% if node['imos_vsftpd']['ssl_enable'] %>
+implicit_ssl=YES
+ssl_tlsv1=YES
+force_local_data_ssl=YES
+force_local_logins_ssl=YES
+rsa_cert_file=/etc/ssl/certs/<% node['imos_vsftpd']['ssl_keypair'] %>.crt
```
lwgordonimos: Should add a sensible default to cookbooks/imos_vsftpd/attributes/default.rb as well so it doesn't explode if the node file doesn't set it.

Ideally, the default should be to use the following, but that will need a slight tweak to the template (since the extension is different for the public key):

```
/etc/ssl/certs/ssl-cert-snakeoil.pem
/etc/ssl/private/ssl-cert-snakeoil.key
```

PR #957, opened 2020-02-25
s3cmd changed to s4cmd

craigrose:
Fixes: https://github.com/aodn/backlog/issues/1348
This PR requires https://github.com/aodn/utilities/pull/194.

Instead of installing s3cmd we install s4cmd.

I have tested this PR (and the backup rock changes) on a vagrant 2-aws-syd node.  Other nodes which should work the same are 6-aws-syd, 3-aws-syd, 18-nec-hob and 3-nec-hob.

Each program uses the same [s3cfg](https://github.com/aodn/chef/blob/master/cookbooks/imos_core/templates/default/s3cfg.erb) file so no changes needed there, but I did include a comments to that effect.

Wherever s3cmd was deployed previously through the imos_backup::default recipe we need to make sure other scripts that depended on it now work correctly with s4cmd.  The only one I could find was [fetch_backup](https://github.com/aodn/chef/blob/master/cookbooks/imos_backup/templates/default/fetch_backup.sh.erb).  This is only currently active on 15-nec-hob, 16-nec-hob and 19-nec-hob.  I have tested the fetch_backup changes on a vagrant 15-nec-hob node.

"backup rock" scripts are also updated to use the s3cmd (see PR https://github.com/aodn/utilities/pull/194).


akashisama:
Looks like a lot of backups failed last night. Presumably relates to this PR getting merged?
image

lwgordonimos:
@akashisama I've fixed this just now by running the `chef_promote` pipeline (since this was the only change).

The reason they failed was because s4cmd wasn't installed by Chef since that change hadn't been promoted.

I'm running the backups manually for those few affected hosts, e.g.

```bash
lwgordon@3-aws-syd:~$ sudo -u backups /mnt/ebs/backups/backup
```

PR #961, opened 2020-03-05
Remove Python 2 dependency installation from data_services recipe

lwgordonimos:


lwgordonimos:
Superseded by https://github.com/aodn/chef/pull/966

PR #968, opened 2020-03-23
Update README.pipeline-box.md

lbesnard:
add information on database connection via talend


lbesnard:
@evacougnon see this PR
this is to avoid the
```
(org.postgresql.util.PSQLException: FATAL: no pg_hba.conf entry for host "1.152.108.141", user "aatams_sattag_qc_dm", database "harvest", SSL off)
``` 
error

PR #972, opened 2020-03-25
Refactor node SSH config generator script

lwgordonimos:
@anguss00 nice one, I crunched a few things down to get rid of some unnecessary logic. 

Also:
1. it is cleaner if it just takes the path to the node file rather than has implicit assumptions about the filesystem
1. it shouldn't write to the home directory, since the point is to keep it self contained in the workspace
1. making the output files contain the node name means it could hypothetically generated multiple nodes without clobbering files from other nodes


anguss00:
Sweet, thanks @lwgordonimos! I'm still having issues with the ssh connect through Jenkins unfortunately but that was before these changes

PR #979, opened 2020-04-08
add libxss1 apt package to jenkins

gsatimos:
@lwgordonimos I need to add this package to Jenkins because it's a dependency of `Cypress`.

To test, I ran up a vagrant 13-aws-syd (base only - no recipes) and installed it manually. Unfortunately I could not fully run up the full 13-aws-syd (not even sure this is possible anymore).

Is it possible to merge and deploy?

re: https://github.com/aodn/backlog/issues/1540


gsatimos:
@lwgordonimos suggested a better approach so closing this

PR #982, opened 2020-05-07
Add AWS profile parameter to Nagios "check_eb" command

lwgordonimos:
The underlying check plugin supports profiles already, so this simply hands the value through from the databag config.


anguss00:
On it

PR #988, opened 2020-05-27
feat(git): ask to update src directories when provisioning

ocehugo:


bin/vagrant-wrapper.sh
```diff
@@ -52,8 +59,17 @@ _clone_git_repo() {
...
         fi
     fi
-
-
+    is_at_tip=$(_is_git_repo_at_tip "$dst")
+    if [ "$is_at_tip" -ne 0 ]; then
+        while true; do
+            read -p "$dst is not updated.Do you wish to update !?\n" res
+            case $res in 
+                [Yy]* ) (cd "$dst" && git checkout master &&  git pull); break;;
```
lwgordonimos, line 67: This is the only thing I think needs to be tested with a few scenarios.

The local branch could be any branch, of any age, with any amount of local changes (committed or uncommitted), so this line needs to handle that somehow sensibly.

Aside from that question, looks pretty good.

bin/vagrant-wrapper.sh
```diff
@@ -52,8 +59,17 @@ _clone_git_repo() {
            (cd $dst ; git checkout $br)
         fi
     fi
-
-
+    is_at_tip=$(_is_git_repo_at_tip "$dst")
+    if [ "$is_at_tip" -ne 0 ]; then
+        while true; do
+            read -p "$dst is not updated.Do you wish to update !?\n" res
```
lbesnard, line 65: it would be better imo if the user wasn't prompted for this, but rahte if it was an argument passed to the main script. I usually run the pipeline on a terminal, and go to another one, come back to it 15min later. If there is a prompt at 5', it is easily missed

lwgordonimos: @lbesnard the solution is up for discussion with POs, since you are the users.

The issue it's trying to solve is that these repos cloned by the script *routinely* become out of date and cause issues, because there's nothing to remind people that they need to be kept up to date.

It needs to be established as early as possible whether these supporting repos are out of date somehow. Otherwise it just fails 5 minutes into the Chef run because of a stale data-services or aodncore or something.

It's pretty unintuitive for Chef to fail because of some repos that a wrapper script cloned, and that a user needs to *remember* (without any prompting) to keep 4(!) external repositories up to date manually.

Clobbering the repos is a recipe for losing "work in progress" so I suggested that being prompted is better than either clobbering, or getting minutes into Chef and then failing with cryptic errors. My concern with a script parameter is that it would have the same problem of just being forgotten and then it just fails like it does currently.

Anyway, that's the context of the issue.

lwgordonimos: > If there is a prompt at 5'

And in response to this, the prompt should be 0.5 seconds after running the script, not 5 minutes :)

bin/vagrant-wrapper.sh
ocehugo: Yeap, I think the best bet is to just stash everything if the user says yes...

bin/vagrant-wrapper.sh
ocehugo: >Otherwise it just fails 5 minutes into the Chef run because of a stale data-services or aodncore or something.

nothing like optimism! I likely spent 30 min trying to debug my problem, then 15+minutes reprovisioning from scratch, bothering you, and finding out that two of your custom repos were too outdated in the src folder to be installed by chef. My fault!? completely, but hey, a pointed finger would be nice.

>My concern with a script parameter is that it would have the same problem of just being forgotten and then it just fails like it does currently.

I Agree. Unless it defaults to prompt, we end up in the same situation. Hence, a flag to disable prompting?

IMO, I think the two most common scenarios/mindsets are "I need to check some stuff - start a new pipeline-box for me with default config", and "I need to test new stuff - start a new pipeline-box for me with this custom repo". Moreover, the latter is usually what I swim towards first (check provisioning/default is working), then get out and change the repo and reprovision the thing.

thus, prompting as default appears the most plausible way.

bin/vagrant-wrapper.sh
lwgordonimos: I think it's probably best to not try and "over automate" this.

If a simple git pull fails, it should probably just say "go and sort out the state of the repository manually then try again". This means changes are never lost, and the original intent is maintained (which is to simply make it immediately *visible* whether something is out of date).

bin/vagrant-wrapper.sh
lwgordonimos: I agree. The default should be to prompt, and if someone wants to bypass the prompt for some reason, they can just supply a "--noninteractive" flag or something.

ocehugo: Okey dokey

lwgordonimos:
As per the discussion, I think it's important that this is primarily informational, and isn't trying to prescribe a particularly workflow each time. That said, I don't mind the idea of some explicit flags to shortcut common, broad use cases.

* Option 1: the user updates the repo(s) manually to the desired state and cooks again
    * Example use case: if there are already existing changes, this allows the user to do any kind of rebasing or merging they may want to do in order to test the changes
* Option 2: the user wants to just cook it "as-is", despite being stale
    * Example use case: the user deliberately rolled back to an old version to test for some reason
* Option 3: the user hasn't pulled the repos for a while, and wants the script to just use the latest master
    * Example use case: the user has no pending changes in any repos, and just wants a shortcut to update everything to the current version

The change I made causes the script to exhibit the following behaviour to support these use cases:

* Option 1: prints a message that the repo is out of date, offers some advice on remediation options, and exits without attempting deployment

* Option 2: the user can set `VAGRANT_ALLOW_OUTDATED_REPOS=true` to ignore the stale state of the repos and deploy the node anyway

* Option 3: the user can set `VAGRANT_CLOBBER_OUTDATED_REPOS=true` to cause the script to reset the repo to the current origin/master before deploying the node

@ocehugo @lbesnard @mhidas does that sound like it gives a reasonable amount of control over all the use cases, while still solving the problem of silently using stale repos?

ocehugo:
My thoughts:
Option 1 - printed diagnostic should be always be printed if repos are not at the upstream tip/synced.
Option 2 - Don't need an option to ignore state, keep it simple, flowing, and following current behaviour (we got a message after all).
Option 3 - should never exist and should be substituted by Option 4:
Option 4 - if ```STASH_modifs=true``` (or alike) option is set, message stashing, stash everything, sync to master, and keep going.

ocehugo:
Humm, thought quickly about it and now I'm verging towards just a message to say (This is outdated/not upstream synch, do you wish to continue?). This is simple, does the job, and if it's the first thing done at the start it will solve the problem.

lwgordonimos:
Take 3 ;)

It will print a warning no matter what, and prompt by default.

To handle it non-interactively, there's a single environment variable, VAGRANT_OUTDATED_REPO_ACTION. This can be set to "clobber, stash or ignore".

PR #993, opened 2020-06-10
Remove jenkins cookbook

anguss00:
https://github.com/aodn/backlog/issues/1776


cookbooks/imos_artifacts/libraries/fetcher_jenkins.rb
```diff
@@ -1,198 +0,0 @@
-#
```
lwgordonimos, line 1: I'd leave the imos_artifacts changes for another PR, and just get rid of the imos_jenkins cookbook in this one, because they are separate things really.

anguss00: No worries, I'll back it out and make another PR

anguss00: @lwgordonimos done, separate PR coming

PR #995, opened 2020-06-11
Remove s3fs-fuse cookbook

anguss00:
https://github.com/aodn/backlog/issues/1776


lwgordonimos:
Couple more things need to happen before we can merge this:

1. the mounts need to be removed from the nodes first
1. there are a bunch of references in `imos_mounts` to be cleaned up
1. there is a reference in Nagios as well (referring to `fuse.sshfs`)

anguss00:
@lwgordonimos with point 3, can we include this in the sshfs removal PR separately?

PR #999, opened 2020-06-15
DOC: update asyncupload with example

lbesnard:


lwgordonimos:
I already did some updates to this doc, see here: https://github.com/aodn/chef/pull/991/files?file-filters%5B%5D=.md#diff-be3fb9a4b54cd15bc6963fcd57def91a

Happy to compromise to make it as clear as possible, but one thing I really don't like is outputting the file directly into $INCOMING_DIR. This is a recipe for errors and invalid manifest files. You should *always* generate the manifest somewhere else first, validate that it is correct and *only then* place it in the incoming directly.

lbesnard:
ah sure sorry for that. feel free to close it then or grab bits you think should be added

lwgordonimos:
Changes made in linked PR.

PR #1005, opened 2020-06-22
S3FS refactor

lwgordonimos:
imos_mounts wasn't really using s3fs-fuse, so this replaces it completely with the standard Ubuntu package, and ensures that imos_mounts can mount is as generically as possible.


ocehugo:
keeping the sweet rolls. Niceee

PR #1021, opened 2020-07-27
WIP: vagrant-aws pipeline box provisioning

anguss00:


.gitignore
```diff
@@ -22,3 +22,5 @@ private-sample/data_bags/passwords/s3_imos_restore.json
 private-sample/nodes/pipeline.json
 
 tmp
+.kitchen/
```
lwgordonimos, line 25: Is this left over from test kitchen prototyping?

chefignore
```diff
@@ -0,0 +1 @@
+.kitchen
```
lwgordonimos, line 1: testkitchen?

vagrant/aws.rb
```diff
@@ -0,0 +1,44 @@
...
+    aws.instance_type = ENV['INSTANCE_TYPE']
+    aws.ami = ENV['SOURCE_AMI']
+    aws.security_groups = ["sg-0a7fce6c", "sg-0db869b77c7b4b418"]
+    aws.subnet_id = ENV['SUBNET_ID']
+    aws.keypair_name = ENV['SSH_KEY']
+    aws.elastic_ip = true
+    aws.associate_public_ip = true
+
+    override.ssh.username = "ec2-user"
+    override.ssh.private_key_path = "/home/grevious/.ssh/vagrant_rsa"
```
lwgordonimos, line 40: This needs a sensible value, not sure what that is just with a cursory glance over this PR, but I think you mentioned the SSH stuff was still TODO.

vagrant/hash.rb
```diff
@@ -0,0 +1,10 @@
+class Hash
+  def slice(*keep_keys)
```
lwgordonimos, line 2: Is this actually used anywhere in this PR?

Vagrantfile
```diff
@@ -54,7 +46,7 @@ Vagrant.configure("2") do |config|
 
   define_node = Proc.new do |node_name|
     config.vm.define node_name, autostart: false do |node|
-      configure_virtualbox_provider node, node_name
+      configure_provider node, node_name
```
lwgordonimos, line 49: It seems like this configures *both* providers every time. Can it be refactored to only configure the provider it's actually about to use?

vagrant/aws.rb
```diff
@@ -0,0 +1,44 @@
...
+  aws_init(ENV['AWS_PROFILE'])
+
+  config.vm.provider :aws do |aws, override|
+    aws.access_key_id = ENV['AWS_ACCESS_KEY_ID']
+    aws.secret_access_key = ENV['AWS_SECRET_ACCESS_KEY']
+    aws.session_token = ENV['AWS_SESSION_TOKEN']
+    aws.region = ENV['REGION']
+    aws.instance_type = ENV['INSTANCE_TYPE']
+    aws.ami = ENV['SOURCE_AMI']
+    aws.security_groups = ["sg-0a7fce6c", "sg-0db869b77c7b4b418"]
```
lwgordonimos, line 33: Need to get rid of any static references to resources somehow. Needs to be totally dynamic/ephemeral.

Rakefile
```diff
@@ -13,3 +13,10 @@ RSpec::Core::RakeTask.new(:spec) do |t, args|
   t.pattern = "cookbooks/*/spec/*_spec.rb"
   t.rspec_opts = "--color"
 end
+
+begin
```
lwgordonimos, line 17: More testkitchen

Vagrantfile
anguss00: I believe that the aws specific provider will only run if --provider=aws is specified - the virtualbox one is the default

lwgordonimos: The provider will, but it looks like this calls both configure methods every time, and then obviously only *uses* one provider? https://github.com/aodn/chef/pull/1021/files#diff-8badd35a3a7b0e15d85fd51aeebca570R1

anguss00: Ah yeah gotcha. I think the stuff within the `config.vm.provider` block will only be executed if that provider is specified, but I see about that entry point. Should have a conditional there

PR #1026, opened 2020-08-03
Add support for optionally encrypted databags

lwgordonimos:
Adds support for autodetection and automatic decryption of encrypted databags.

* Requires that the key be sourced and copied to `chef/encrypted_data_bag_secret` (but not committed!)
* Uses the Chef DSL method `data_bag_item` to handle the magical and transparent autodetection of encrypted data bags
* Allows encrypted databags to be managed as described at https://deangrant.wordpress.com/2015/11/08/chef-creating-encrypted-data-bags-with-knife-solo/
* Uses fork of chef-solo-search from https://github.com/edelight/chef-solo-search/pull/71, to support searching encrypted data bags (IMPORTANT NOTE: forking is bad obviously, but chef-solo will soon be replaced by chef-zero, so this fork should be short-lived)


lwgordonimos:
### Testing

```bash
# encrypt some data bag items using the utility script and the dummy key (don't commit these obviously)
$ ls -1 private-sample/data_bags/passwords/*.json | while read db; do bin/encrypt_databag.rb --path $db --key ./encrypted_data_bag_secret; done
$ ls -1 private-sample/data_bags/users/*.json | while read db; do bin/encrypt_databag.rb --path $db --key ./encrypted_data_bag_secret; done

# deploy pipeline box to verify it cooks as normal
$ bin/pipeline-box.sh
```

PR #1030, opened 2020-08-13
vagrant aws provisioning

anguss00:
https://github.com/aodn/backlog/issues/1919


Vagrantfile
```diff
@@ -45,13 +33,17 @@ Vagrant.configure('2') do |config|
     f.chomp(node_file_extension)
   end
 
-  config.vm.provision 'shell', inline: 'test `find /var/cache -maxdepth 1 -name apt -mtime -1 | wc -l` -eq 0 && sudo apt-get update; true'
+  config.vm.provision "shell", inline: "sudo systemctl disable apt-daily.service && sudo systemctl disable apt-daily.timer"
+  config.vm.provision "shell", inline: "sudo rm /var/lib/dpkg/lock && sudo dpkg --configure -a && sudo apt-get install -f"
```
lwgordonimos, line 37: This stuff should be all handled during image creation, and not repeated during every Vagrant deployment. As discussed, I think any issues around dpkg locking should have the root cause dealt with, because the more bash that gets run here, the longer it takes and the more fragile it is.

vagrant/aws.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  response = {
+      :access_key_id => sts.config.credentials.credentials.access_key_id,
+      :secret_access_key => sts.config.credentials.credentials.secret_access_key,
+      :session_token => sts.config.credentials.credentials.session_token
+  }
+  # # # cloudformation = Aws::CloudFormation::Client.new(credentials: credentials, region: region)
+
+  # # TODO: discover from CF outputs, if required ?
+  response[:subnet_id] = 'subnet-19956d7e'
```
lwgordonimos, line 14: We definitely need to sort out this TODO, to get rid of account specific hard coded values.

PR #1036, opened 2020-08-26
Chef 15 support

lwgordonimos:
I have been testing on 15.13.8, since that is the version in the latest stable ChefDK.

Summary of changes required:

* General cookbook version compatibility with Chef
~~* `params` resource parameter renamed due to naming clash~~
~~* `node.set` removed, replaced with node.override~~
* java cookbook changed to use resource model
* Ohai now treats 'passwd' plugin as optional and not run by default (TODO: get plugin to run again)
~~* search method can return different types, so helper class needs to explicitly coerce return value~~
~~* user resource manage_home argument syntax change~~
~~* supervisor cookbook forked and patched for Chef > 12 support~~
* newer Chef versions require acceptance of license before use


lwgordonimos:
### TODO
- [ ] split this into two PRs, one which can be merged and work on Chef 12.4.3, and a second which is changes to merge only when actually upgrading.

lwgordonimos:
Superseded by https://github.com/aodn/chef/pull/1067

PR #1037, opened 2020-08-26
as a PO I want to use locate & ag so I can search for (within) files …

ocehugo:
…(logs) efficiently

@lwgordonimos - mlocate need config since we don't want inspection over /mnt/ebs/imos-data and alikes

legit?


lbesnard:
I don;t think those kind of tools should necessary be installed system wide. You can always install them in your home directory (lin

lwgordonimos:
Couple of things.

This Dockerfile is only used by Jenkins to run the test suite. Packages to be installed by data-services cookbook to create the "standard PO environment" are defined at https://github.com/aodn/chef/blob/master/cookbooks/imos_po/attributes/data_services.rb#L42

I kind of agree with Loz, this recipe already installed a huge amount of things (each of which makes cooking take longer for everything, consumes space (both prod and dev environments)). So we need to be really careful to not add packages without a good business case (i.e. it contributes positively to the workflow for all POs). There shouldn't really be very many variables in terms of *where* files are as part of the pipeline workflow, so what is being searched for? What problem does it solve that `find` doesn't, when it comes to searching the pipeline directories?

ocehugo:
@lwgordonimos - I thought I answered this thing.

I'm happy that I don't need to bring my own `find` to my home folder! Jokes apart, my point of suggesting these is that if a tool is a better fit, why not make it available to everyone?  I'm happy to bring my own software stack to my home, but should I do it for ubiquitous tools?

`mlocate` reduce the load on the disk, makes searching files way faster, and avoid repeated calls to find when the recursive search bites you. I'm also not the type of guy to keep in memory every file name or every location of a particular feature (pipeline config, log location, file location, etc). `mlocate` is the local google. 

`ag` help a lot because our logs are not rotated and huge. It also permits a much better search experience by filtering the search query by filetype. I'm happy to keep using grep but the current experience is not snappy enough - we all love a good and fast text query anyway.

Also, since I'm here, I noticed yesterday that `rename` in prod/rc is from a perl package, instead of being from the kernel/utils sources. I think it's better to install the `util-linux` package, since it's more "standard" (reducing perl things is always a good thing).


>this recipe already installed a huge amount of things (each of which makes cooking take longer for everything, consumes space (both prod and dev environments)).

All of the above should take less than 5mb of local space (~4mb for util-linux). This will take what, 1s more for deployment?

lwgordonimos:
1. I don't really have a problem with `silversearcher-ag`, as a small standalone utility, the PR to add that is here: https://github.com/aodn/chef/pull/1069
2. `mlocate` introduces the database, cronjobs (i.e. isn't just disk space, is moving parts and CPU cycles) and needing Chef work to manage exclusions for servers with >2TB local disk mounted and the notoriously flaky S3FS mounted as well. It just isn't a good fit for this type of batch processing server without a more compelling reason.
3. Reducing Perl utilities is not a priority, or any real advantage at all when they are just OS utilities. Not keen to change things without a valid reason, I have more than enough important work to do.

ocehugo:
Thanks

lwgordonimos:
silversearcher-ag has been added in: https://github.com/aodn/chef/pull/1069

PR #1041, opened 2020-09-01
Data-services databag refactor

anguss00:


lwgordonimos:
I feel like this has gone down a Ruby rabbit hole and away from Chef DSL, and there's a lot of broken abstractions as a result ;)

https://github.com/aodn/chef/pull/1042

Have a look here at adding the path at converge time. Basically, as soon as data-services is confirmed to exist, it adds the path, and any subsequent databag loading code will transparently use it.

The only thing to do do is then make talend a) formally depend on imos_po cookbook and b) update that loop to move that loop out of compile time, but it really needs to use a Chef resource model to avoid all the Ruby weirdness.

cookbooks/talend/recipes/default.rb
```diff
@@ -33,47 +29,10 @@ def get_job_common_name(job_name, data_bag_item)
...
-  trigger = {}
-  trigger['cron'] = cron
-  trigger['events'] = events
-
-  job_common_name = get_job_common_name(job_name, bag_item)
-
-  Chef::Application.fatal!("Data bag '' does not have artifact_uri defined") unless bag_item['artifact_uri']
 
-  job_id = "#{job_name}-#{job_name}"
+  bag_item = Talend::JobHelper.get_new_talend_job_config(job_name, node)
```
lwgordonimos, line 33: This is probably a useful refactor at the Ruby class level, but I wonder if we couldn't move the method call (and importantly the databag load) into the `talend_job` resource, to get it out of compile phase and into the converge phase?

cookbooks/imos_po/libraries/pipeline_utils.rb
```diff
@@ -11,7 +11,18 @@ module ImosPo
   class PipelineUtils
     # static utility class for pipeline supporting code
 
-    def self.get_new_watch(watch, strip_notify_lists = true)
+    def self.set_databag_path(node)
```
lwgordonimos, line 14: See my comment on the PR, I think this should be handled already by Chef, and not have databag stuff being leaked into these helper classes.

anguss00:
@lwgordon OK no worries, that looks heaps cleaner

lwgordonimos (changes requested):

cookbooks/talend/recipes/default.rb
anguss00: Isn't this following the refactor of watch config though? That's what I based it off, refactoring the databag load into a library function rather than a resource. The resource stuff I found hard to reconcile as there's subsequent talend_job calls that use parts of the returned databag item

anguss00: ah sorry, just re-read the original comment - makes sense!

lwgordonimos: The trouble is that this is just executing Ruby code in the recipe body still.

I can't see why you can't just supply `job_name` as a parameter to `talend_job` so that it can load the databag item when the resource executes? It seems like it's only ever giving the bag_item values to the resource anyway, so just let the resource load it.

anguss00: totally - yeah I understood above once I re-read hah

lwgordonimos: Actually, it's already a parameter so it shouldn't be much more than refactoring this method call into the resource provider code:

https://github.com/aodn/chef/blob/master/cookbooks/talend/providers/job.rb#L29

cookbooks/imos_po/attributes/data_services.rb
```diff
@@ -24,6 +24,7 @@
 
 # env file for data-services repository
 default['imos_po']['data_services']['env'] = ::File.join(node['imos_po']['data_services']['dir'], 'env')
+default['imos_po']['data_services']['data_bag_dir'] =  ::File.join(node['imos_po']['data_services']['dir'], 'data_bags')
```
lwgordonimos, line 27: Can get rid of this now

cookbooks/imos_po/libraries/pipeline_utils.rb
```diff
@@ -34,16 +34,5 @@ def self.get_new_watch(watch, strip_notify_lists = true)
 
       data_bag_hash
     end
-
```
lwgordonimos, line 37: This is unrelated code, so should leave this as is

cookbooks/talend/providers/job.rb
```diff
@@ -54,15 +56,16 @@ def load_current_resource
...
     variables(
       params: evaluated_job_parameters,
-      delimiter: new_resource.delimiter
+      delimiter: job_config['delimiter']
     )
   end
 
   # process harvester specific resources
-  unless new_resource.harvest_resources.nil?
+  harvest_resources = job_config['resources']
```
lwgordonimos, line 64: Don't need to alias this, just use `job_config['resources']`

cookbooks/talend/providers/job.rb
```diff
@@ -129,11 +133,13 @@ def load_current_resource
 
 # Build wrapper script for job
 action :schedule do
+  job_config = Talend::JobHelper.get_new_talend_job_config(@data_bag_item_name, node)
   # Change privileges of artifact job script
   ::FileUtils.chmod(00755, Talend::JobHelper.job_script_path(new_resource))
   ::FileUtils.chown(new_resource.owner, new_resource.group, Talend::JobHelper.job_script_path(new_resource))
 
-  if new_resource.trigger['events'].empty?
+  trigger = job_config['trigger']
```
lwgordonimos, line 141: No need to alias this, just use `job_config['trigger']`

cookbooks/talend/recipes/default.rb
```diff
@@ -120,7 +71,6 @@ def get_job_common_name(job_name, data_bag_item)
 jobs_to_cleanup.uniq.each do |job_id|
   talend_job job_id do
     action      :remove
-    common_name job_id
```
lwgordonimos, line 123: Don't need to remove this

lwgordonimos (changes requested):

cookbooks/talend/libraries/job_helper.rb
```diff
@@ -163,5 +163,68 @@ def self.hash_as_json_to_file(file, config)
...
+      job_id = "#{job_name}-#{job_name}"
+
+      bag_item['job_common_name'] = job_common_name
+      bag_item['job_parameters'] = job_parameters
+      bag_item['trigger'] = trigger
+
+      bag_item
+    end
+
+    def self.get_log_file_for_job(job_name, node)
```
lwgordonimos, line 222: Is this method used anywhere?

private-sample/data_bags/imos_po_watches/ANMN_AM.json
```diff
@@ -1,19 +0,0 @@
-{
```
lwgordonimos, line 1: I'd leave these databags in this PR, because this is just about the code feature, and this is really config changes (which should be considered separately).

lwgordonimos: Sorry, being pedantic, but these still have whitespace changes in these databags which shouldn't be in the PR

anguss00: No sorry my bad - I copied from current master and something screwed up with the tab spaces

lwgordonimos (approved):

PR #1045, opened 2020-09-09
update vagrant dependencies

leonardolaiolo:


lwgordonimos:
We only support the Virtualbox provider for Vagrant (not libvirt), because it is a single cross-platform option, and also that is the only thing we build the base images for.

I'm wondering what this one was for? I double-checked that I don't have libvirt-dev installed, so it's necessary for running Vagrant on Virtualbox.

ocehugo:
@lwgordonimos my bad, I think this dependency creep was because of some specific version build while we were trying to install the bit...

I still can't believe that ubuntu maintain a broken vagrant/berkshelf combination on apt...

PR #1063, opened 2020-10-15
Apache2 php fixes

anguss00:


cookbooks/imos_apache2/recipes/php.rb
```diff
@@ -9,12 +9,16 @@
...
 template 'php7.conf' do
   extend Apache2::Cookbook::Helpers
-  path lazy { "#{apache_dir}/mods-available/#{apache_mod_php_modulename}.conf" }
+  path lazy { "#{apache_dir}/mods-available/#{node['imos_apache2']['php']['module_name']}.conf" }
   source 'default/php7.conf.erb'
   mode '0644'
   notifies :reload, 'service[apache2]', :delayed
 end
 
+apache2_module node['imos_apache2']['php']['module_name'] do
```
lwgordonimos, line 18: Given this is a workaround, we should add a comment here to that effect.

Also, if this fully supersedes the `apache2_mod_php` resource below, we should get rid of that line.

cookbooks/imos_apache2/recipes/php.rb
```diff
@@ -9,12 +9,16 @@
 
 template 'php7.conf' do
   extend Apache2::Cookbook::Helpers
-  path lazy { "#{apache_dir}/mods-available/#{apache_mod_php_modulename}.conf" }
+  path lazy { "#{apache_dir}/mods-available/#{node['imos_apache2']['php']['module_name']}.conf" }
```
lwgordonimos, line 12: This should either be hardcoded to match the `source` line or they should both be derived from this attribute. It doesn't make sense to have one as a variable and one as hardcoded.

lwgordonimos (changes requested):

anguss00:
@lwgordonimos updated

PR #1065, opened 2020-10-28
Consolidate serverless monitoring templates

lwgordonimos:
This is a prerequisite for refactoring the imos_nagios cookbook, which will require reducing the number of these "raw" templates as much as possible.


anguss00:
@lwgordonimos tests OK. Just have a question regarding the aws-wps prod bucket monitoring

anguss00 (approved):

PR #1070, opened 2020-11-18
create new schema for NRMN

ana-berger:
Please, could you check if that is correct for creating a new schema?


mhidas (approved):
Looks fine to me.
However, I'm not sure this is actually going to have any effect. Config in the chef repo is only used in the pipeline box, but even that box uses the database in the stack (configured elsewhere!). So I don't think there's any point adding these schemas in chef anymore. 

Correct me if I'm wrong @lwgordonimos

mhidas:
@lwgordonimos Are we ready to start adding this sort of config in data-services instead??

mhidas:
@ana-berger For RC & prod, you'll need to put the config in chef-private (see https://github.com/aodn/chef-private/pull/3583 for example)
For the po stack, the same config goes in the [cloud-deploy](https://github.com/aodn/cloud-deploy/tree/master/sample-config/data_bags) repo, I think.

ana-berger:
@mhidas yes I knew I need to add config in chef-private as well. Going through it right now.

ana-berger:
> For the po stack, the same config goes in the [cloud-deploy](https://github.com/aodn/cloud-deploy/tree/master/sample-config/data_bags) repo, I think.
So, for the stack to have the schema we need to add it in cloud-deploy? Is that ok if I send a PR then to do that or is it something for the devs? @lwgordonimos

lwgordonimos:
> Are we ready to start adding this sort of config in data-services instead??
Not yet unfortunately, there were questions to be answered around database credentials first.

> Is that ok if I send a PR then to do that or is it something for the devs?

Just for completeness, the best thing until we sort this mess out a bit is to add it here as you need it, and I'll just replicate it over to cloud-deploy. Sorry it's pretty annoying, but kind of in a transitional phase...

ana-berger:
All good @lwgordonimos, I got it. Thank you for the clarification!

